PR #10, opened 2021-02-15
Move copy-pastable example to bevy_egui 0.2

ssbr:


ssbr:
IDK if there's a way to automate this, I end up sending these kinds of PRs to a lot of Rust projects. Anyway, found out the hard way I was building on an out of date version because I had copy-pasted the readme dep. (Unavailability of `clicked()` was the big indicator).

I like the CI setup btw. :)

mvlabat:
Oh, indeed, thanks for the PR. :)

I've recently stumbled upon this: https://crates.io/crates/version-sync. Looks like it's time to add it to the deps.

PR #3, opened 2021-01-19
Directly use EventReader

Weasy666:
Use EventReader directly, which was made possible by https://github.com/bevyengine/bevy/pull/1244.

This should not be merged before the release of Bevy 0.5, as it needs at least commit https://github.com/bevyengine/bevy/commit/a880b5450829411b0248e47cc066915234371d86 to work.

One test is failing, but i am not sure if it has something to do with my changes. I tested this in one of my project and it works just fine.

```
running 2 tests
test src\lib.rs - EguiSettings::scale_factor (line 100) ... ok
test src\lib.rs - (line 23) ... FAILED

failures:

---- src\lib.rs - (line 23) stdout ----
Test executable failed (exit code 101).

stderr:
thread 'Compute Task Pool (2)' panicked at 'assertion failed: pixels_per_point > 0.0', C:\Users\...\.cargo\registry\src\github.com-1ecc6299db9ec823\epaint-0.8.0\src\text\font.rs:76:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'task has failed', C:\Users\...\.cargo\registry\src\github.com-1ecc6299db9ec823\async-task-4.0.3\src\task.rs:368:45



failures:
    src\lib.rs - (line 23)

test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.15s
```

PR #12, opened 2021-02-20
update to bevy 0.5

jakobhellermann:
This shouldn't be merged yet, I'm just putting this up as a draft PR so that anyone who wants to update to bevy main can see that the work is already done.

includes #3

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,8 +16,8 @@ manage_clipboard = ["clipboard", "thread_local"]
 open_url = ["webbrowser"]
 
 [dependencies]
-bevy = { version = "0.4", default-features = false, features = ["render"] }
-bevy_winit = "0.4.0"
+bevy = { git = "https://github.com/bevyengine/bevy", default-features = false, features = ["render"] }
+bevy_winit = { git = "https://github.com/bevyengine/bevy" }
 egui = "0.9.0"
 webbrowser = { version = "0.5.5", optional = true }
 winit = { version = "0.24.0", features = ["x11"], default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ license = "MIT"
 edition = "2018"
 repository = "https://github.com/mvlabat/bevy_egui"
 exclude = ["assets/**/*", ".github/**/*"]
+resolver = "2"
 
 [package.metadata.docs.rs]
 no-default-features = true # clipboard crate fails the build currently https://github.com/rust-lang/docs.rs/issues/695
@@ -31,12 +32,5 @@ thread_local = { version = "1.1.0", optional = true }
 [dev-dependencies]
 version-sync = "0.9.2"
 
-[[example]]
-name = "ui"
-path = "examples/ui.rs"
-required-features = ["bevy/x11", "bevy/png", "bevy/bevy_wgpu"]
-
-[[example]]
-name = "simple"
-path = "examples/simple.rs"
-required-features = ["bevy/x11", "bevy/bevy_wgpu"]
+[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
+bevy = { git = "https://github.com/bevyengine/bevy", branch = "main", default-features = false, features = ["bevy_wgpu", "x11", "png"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,10 +26,13 @@ winit = { version = "0.24.0", features = ["x11"], default-features = false }
 clipboard = { version = "0.5.0", optional = true }
 thread_local = { version = "1.1.0", optional = true }
 
+[dev-dependencies]
+bevy = { git = "https://github.com/bevyengine/bevy", default-features = false, features = ["x11", "bevy_wgpu"] }
+
 [[example]]
 name = "ui"
 path = "examples/ui.rs"
-required-features = ["bevy/x11", "bevy/png", "bevy/bevy_wgpu"]
+required-features = ["bevy/png"]
 
 [package.metadata.docs.rs]
 no-default-features = true # clipboard crate fails the build currently https://github.com/rust-lang/docs.rs/issues/695
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,8 +16,7 @@ manage_clipboard = ["clipboard", "thread_local"]
 open_url = ["webbrowser"]
 
 [dependencies]
-bevy = { git = "https://github.com/bevyengine/bevy", default-features = false, features = ["render"] }
-bevy_winit = { git = "https://github.com/bevyengine/bevy" }
+bevy = { git = "https://github.com/bevyengine/bevy", default-features = false, features = ["render", "bevy_winit"] }
 egui = "0.9.0"
 webbrowser = { version = "0.5.5", optional = true }
 winit = { version = "0.24.0", features = ["x11"], default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ open_url = ["webbrowser"]
 
 [dependencies]
 bevy = { version = "0.5", default-features = false, features = ["render", "bevy_winit"] }
-egui = "0.10.0"
+egui = "0.11.0"
 webbrowser = { version = "0.5.5", optional = true }
 winit = { version = "0.24.0", features = ["x11"], default-features = false }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ manage_clipboard = ["clipboard", "thread_local"]
 open_url = ["webbrowser"]
 
 [dependencies]
-bevy = { git = "https://github.com/bevyengine/bevy", default-features = false, features = ["render", "bevy_winit"] }
+bevy = { git = "https://github.com/bevyengine/bevy", branch = "main", default-features = false, features = ["render", "bevy_winit"] }
 egui = "0.9.0"
 webbrowser = { version = "0.5.5", optional = true }
 winit = { version = "0.24.0", features = ["x11"], default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ manage_clipboard = ["clipboard", "thread_local"]
 open_url = ["webbrowser"]
 
 [dependencies]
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -358,6 +355,10 @@ impl Plugin for EguiPlugin {
         render_graph
             .add_node_edge(base::node::MAIN_PASS, node::EGUI_PASS)
             .unwrap();
+        render_graph
+            // hardcodede until https://github.com/bevyengine/bevy/pull/1464 is merged
+            .add_node_edge("ui_pass", node::EGUI_PASS)
+            .unwrap();
```
mvlabat, line 365: I'm curious whether this works if a user doesn't have the UI plugin enabled. Probably, we'll have to add this node conditionally based on whether such a node exists or the plugin is enabled. I'm not sure there's even API for checking it though. 🤔

jakobhellermann: Now it only gets added [if the ui pass node exists](https://github.com/mvlabat/bevy_egui/pull/12/commits/caa7eacac1126a9cc670136131a16c7456fdb6ec)

src/lib.rs
```diff
@@ -297,39 +298,46 @@ pub mod node {
...
+            CoreStage::PreUpdate,
+            begin_frame
+                .system()
+                .label(EguiSystem::BeginFrame)
+                .after(EguiSystem::ProcessInput),
+        );
+        app.add_system_to_stage(
+            EguiStage::UiFrameEnd,
+            process_output.system().label(EguiSystem::ProcessOutput),
+        );
```
mvlabat, line 340: I think I'd like to preserve some of the stages. You can see that some of them were added, but no library systems used them. I designed those stages to leave users a possibility to hook into them and change or read from the Input/Output state, so that they can re-implement clipboard functionality, opening URLs, etc.

(I know, I should have documented my intentions here better, on which I'll definitely work more.)

I haven't looked much into system labels yet, but if the same is achievable by making `EguiSystem` public, for instance, I'll be glad to keep the refactoring as is.

jakobhellermann: I made the `EguiSystem` is public, people can now have dependencies like `system.after(EguiSystem::ProcessInput).before(EguiSystem::BeginFrame)`, so hooking into that should work

Cargo.toml
```diff
@@ -19,8 +20,7 @@ manage_clipboard = ["clipboard", "thread_local"]
 open_url = ["webbrowser"]
 
 [dependencies]
-bevy = { version = "0.4", default-features = false, features = ["render"] }
-bevy_winit = "0.4.0"
+bevy = { git = "https://github.com/bevyengine/bevy", branch = "main", default-features = false, features = ["render", "bevy_winit"] }
```
piedoom, line 23: Would we still want bevy master, or bevy 0.5 now that it's released?

jakobhellermann: Yep, I've changed it now

Cargo.toml
```diff
@@ -32,7 +32,5 @@ thread_local = { version = "1.1.0", optional = true }
 [dev-dependencies]
 version-sync = "0.9.2"
 
-[[example]]
-name = "ui"
-path = "examples/ui.rs"
-required-features = ["bevy/x11", "bevy/png", "bevy/bevy_wgpu"]
+[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
+bevy = { version = "0.5", default-features = false, features = ["bevy_wgpu", "x11", "png"] }
```
mvlabat, line 36: Oh, this is great! I guess it also lets us tidy this feature matrix a bit
https://github.com/mvlabat/bevy_egui/blob/main/.github/workflows/check.yml#L34

src/transform_node.rs
```diff
@@ -1,11 +1,11 @@
 use crate::{EguiSettings, WindowSize, EGUI_TRANSFORM_RESOURCE_BINDING_NAME};
 use bevy::{
     core::AsBytes,
-    ecs::{Commands, IntoSystem, Local, Res, ResMut, Resources, System, World},
+    prelude::*,
```
mvlabat, line 4: I'd like to avoid this one as well

mvlabat:
I believe that at the moment `bevy_egui` doesn't work with `bevy_webgl2` since it's not yet upgraded to 0.5. I'd like to hold off the release of 0.5 support until `bevy_webgl2` is upgraded, but I'm fine with merging this to master earlier. So at least it'll be possible for users to point dependency resolution to this repo instead of the fork.

jakobhellermann:
Alright, bevy_webgl2 shouldn't take too long to release, IIRC it is already ported to bevy main and just needs to be published.

mvlabat:
Thanks for the egui upgrade! I'm going to test and merge this tomorrow :)

mvlabat:
@jakobhellermann I believe there's a regression with users' textures rendering. See the `ui` example, now the Bevy logo isn't displayed (it should be in the left panel, like on the screenshot in the readme). I'm also not able to work on any Bevy 0.5 project after wgpu was upgraded to 0.7. I'm on Windows, here every project crashes for me on startup now: https://github.com/gfx-rs/wgpu/issues/1317 :(

I can test things on my old Macbook, but it takes forever to compile... So I won't be able to help with debugging very actively.

mvlabat:
Ok, now it doesn't crash but hangs. :)

src/systems.rs
```diff
@@ -204,7 +211,13 @@ pub fn process_output(
 
     if let Some(window) = windows.get_primary() {
         if let Some(winit_window) = winit_windows.get_window(window.id()) {
-            winit_window.set_cursor_icon(egui_to_winit_cursor_icon(output.cursor_icon));
+            match egui_to_winit_cursor_icon(output.cursor_icon) {
+                None => winit_window.set_cursor_visible(false),
+                Some(icon) => {
+                    winit_window.set_cursor_visible(true);
```
mvlabat, line 217: This line causes a dead-lock on my system O_o

jakobhellermann:
The texture ui seems to have been broken by the egui 0.11 upgrade, I'll look into it.

I also noticed that sometimes minimizing an egui window will lead to a
```
wgpu error: Validation Error

Caused by:
    In a RenderPass
      note: encoder = `<CommandBuffer-(1, 114, Vulkan)>`
    In a set_scissor_rect command
    Invalid ScissorRect parameters
```

src/systems.rs
jakobhellermann: Weird. The `egui::CursorIcon` got a new variant `None` which isn't present in `winit::window::CursorIcon`.

Maybe setting the cursor to visible or invisible every frame isn't the way to go :sweat_smile:

jakobhellermann: I pushed a commit which only sets the cursor visibility if it is changed. Can you try if that fixes it?

mvlabat: Btw, do we need it or do we need to do this in such a way? I tried to look up how Bevy manages window cursors and found this:
https://github.com/bevyengine/bevy/blob/100e516014d803392a81a4e001d3e6ddf5cca414/crates/bevy_winit/src/lib.rs#L115-L118
Which led me to this: https://docs.rs/bevy_window/0.5.0/bevy_window/struct.Window.html#method.set_cursor_visibility

One of my other concerns is that a user might have `WindowDescriptor::cursor_visible` configured to false, while this code may override this. I'm thinking that we should probably only manage cursor icons here and let users control cursor visibility. What do you think?

I guess, if `egui_to_winit_cursor_icon` doesn't return anything, we can either stick to the default cursor or don't change it at all.

jakobhellermann: Done. Now the cursor icon doesnt change if `None` is returned.

jakobhellermann:
The ui is broken since https://github.com/emilk/egui/commit/589bae121116bcf5ae76f46313934c88624fbc81.

Removing the left to right layout in https://github.com/mvlabat/bevy_egui/blob/e5d083fa7ca8fb19884541d7e627e51566824838/examples/ui.rs#L64 fixes the problem.

jakobhellermann:
I opened an issue: https://github.com/emilk/egui/issues/289

mvlabat:
> The ui is broken since [emilk/egui@589bae1](https://github.com/emilk/egui/commit/589bae121116bcf5ae76f46313934c88624fbc81).
> 
> Removing the left to right layout in
> 
> https://github.com/mvlabat/bevy_egui/blob/e5d083fa7ca8fb19884541d7e627e51566824838/examples/ui.rs#L64
> 
> fixes the problem.

It seems like doing
```
ui.allocate_space(egui::Vec2::new(1.0, 100.0));
ui.horizontal(|ui| {
    load = ui.button("Load").clicked();
    invert = ui.button("Invert").clicked();
    remove = ui.button("Remove").clicked();
});
```

does the trick for us as well.

mvlabat:
Looks great! Thanks again for your work on this

jakobhellermann:
> The texture ui seems to have been broken by the egui 0.11 upgrade, I'll look into it.
> 
> I also noticed that sometimes minimizing an egui window will lead to a
> 
> ```
> wgpu error: Validation Error
> 
> Caused by:
>     In a RenderPass
>       note: encoder = `<CommandBuffer-(1, 114, Vulkan)>`
>     In a set_scissor_rect command
>     Invalid ScissorRect parameters
> ```

This is reproducible using
```rust
    egui::Window::new("Window")
        .scroll(true)
        .show(&egui_context.ctx, |ui| {
            ui.label("Windows can be moved by dragging them.");
            ui.label("They are automatically sized based on contents.");
            ui.label("You can turn on resizing and scrolling if you like.");
            ui.label("You would normally chose either panels OR windows.");
        });
```

jakobhellermann:
Can I also suggest one change before releasing, add a `egui_context.ctx()` function and make the `ctx` field private. That way #13 can be done without a breaking change.

mvlabat:
Yeah, I even hope we can merge #13 before the major release. :)

mvlabat:
> This is reproducible using
> 
> ```rust
>     egui::Window::new("Window")
>         .scroll(true)
>         .show(&egui_context.ctx, |ui| {
>             ui.label("Windows can be moved by dragging them.");
>             ui.label("They are automatically sized based on contents.");
>             ui.label("You can turn on resizing and scrolling if you like.");
>             ui.label("You would normally chose either panels OR windows.");
>         });
> ```

I've just stumbled upon this as well. I believe we should file an issue to `egui` about this. It sometimes returns negatives for clipping zone height.

PR #13, opened 2021-04-03
Support for multiple windows

jakobhellermann:
builds on top of #12

[Diff only to #12](https://github.com/jakobhellermann/bevy_egui/compare/bevy-0.5...jakobhellermann:multiple-windows)

Changes:
- `&egui_context.ctx` -> `egui_context.ctx()`
- new `egui_context.ctx_for_window(window_id)` for drawing to another window
- [`egui::setup_pipeline`](https://github.com/jakobhellermann/bevy_egui/blob/c30710c18daa2abfb93d0bd6ad5f547c297d9a56/src/lib.rs#L429) and `egui::RenderGraphConfig` which needs to get run for additional windows
- `Egui{Shapes, Output, Input}` -> `HashMap<WindowId, Egui{Shapes, Output, Input}>`

Let me know what you think :smile:

PR #14, opened 2021-04-10
Multiple windows support

mvlabat:
Based on #13 

Issues to fix:
- [x] When a new window is spawned, it acquires the focus, at least on Windows (TODO: test other platforms), but the active context is still the default window. That results in events coming from the second window being consumed by the default context.
- [x] Clipboard can paste into several fields at the same time if they are all active in different windows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.4.0] - 10-Apr-2021
+
+Huge thanks to @jakobhellermann and @Weasy666 for contributing to this release!
+
+### Added
+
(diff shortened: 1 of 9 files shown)


mvlabat:
@jakobhellermann I've extended the two windows example a bit and found a couple of minor bugs. I'm looking into them right now, but if you already have an idea on how to fix those, that will help. :)

jakobhellermann:
> @jakobhellermann I've extended the two windows example a bit and found a couple of minor bugs. I'm looking into them right now, but if you already have an idea on how to fix those, that will help. :)

What bugs are there? Running locally seems to work for me.

mvlabat:
@jakobhellermann sorry, I forgot to mention that I've described them in the PR body. :) The clipboard issue is now fixed, now I'm trying to figure out what to do with the invalid initial focus.

jakobhellermann:
On linux the focus doesn't seem to be an issue. The second window gets focused as well, but tab presses etc. get correctly handled in the second window.

mvlabat:
Yeah, it looks Windows specific... I wasn't able to reproduce it on my Macbook either.

Issue #15, opened 2021-04-13
Crash when minimising the "Window" widget


stephanbuys:
Hi,

When I run the `ui.rs` example and I "minimize" the "Window" widget I consistently get the following crash ( https://github.com/mvlabat/bevy_egui/blob/main/examples/ui.rs#L110):

```
wgpu error: Validation Error

Caused by:
    In a RenderPass
      note: encoder = `<CommandBuffer-(1, 284, Metal)>`
    In a set_scissor_rect command
    Invalid ScissorRect parameters


thread 'main' panicked at 'Handling wgpu errors as fatal by default', /.../.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.7.1/src/backend/direct.rs:1896:5
stack backtrace:
  11: <bevy_egui::egui_node::EguiNode as bevy_render::render_graph::node::Node>::update
             at ./src/egui_node.rs:266:9
          at ./examples/ui.rs:7:5
```

mvlabat:
Hi @stephanbuys! Thanks for creating the issue. Yeah, I was able to reproduce it while working on the 0.4.0 release as well. I believe that Egui started sending `ScissorRect` with negative height, which is problematic for some graphics backends. I think that this is a problem on the Egui side, though I'm not 100% sure about this. I was going to reach out to the Egui author about this, but I didn't have a chance yet.

If you'd like to investigate this deeper and probably report this problem in the Egui repo, I'd appreciate this. But anyway, I was going to do that myself at the weekend.

stephanbuys:
Ok, so I "patched" the file to see if I could do a hotfix as follows (https://github.com/mvlabat/bevy_egui/blob/main/src/egui_node.rs#L313)

```
                    let scale_factor = window_size.scale_factor * egui_settings.scale_factor as f32;

                    let (x,y,w,mut h) = (
                        (draw_command.clipping_zone.min.x * scale_factor) as u32,
                        (draw_command.clipping_zone.min.y * scale_factor) as u32,
                        (draw_command.clipping_zone.width() * scale_factor) as u32,
                        (draw_command.clipping_zone.height() * scale_factor) as u32
                        );

                    log::info!("{:?}, {:?}, {:?}, {:?}", x,y,w,h);

                    if h == 0 {
                        h = 1;
                    }

                    render_pass.set_scissor_rect(
                        x,y,w,h
                    );
```

[...]

PR #16, opened 2021-04-14
Do not attempt to render widgets with no height or width.

stephanbuys:
This is a suggested fix for #15 

The idea is that if a widget would be either 0 in height or 0 in width (impossible to render), we should do just that, skip attempting to render it.

### DIFF
--- a/src/egui_node.rs
+++ b/src/egui_node.rs
@@ -311,7 +311,7 @@ impl Node for EguiNode {
 
                     let scale_factor = window_size.scale_factor * egui_settings.scale_factor as f32;
 
-                    let (x, y, w, mut h) = (
+                    let (x, y, w, h) = (
                         (draw_command.clipping_zone.min.x * scale_factor) as u32,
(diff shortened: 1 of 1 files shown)


mvlabat:
Hi! Thanks for the PR.

Regarding your comment in the issue:
> I submitted my suggested fix as PR #16 I don't think it's a egui issue, I wasn't able to induce any negative numbers.

I was actually able to reproduce it. I was just rage-clicking the minimize button in the `ui` example, while `dbg!`-printing scissor params, and I've got this:

```
[src\egui_node.rs:313] (draw_command.clipping_zone.min.x, draw_command.clipping_zone.min.y,
 draw_command.clipping_zone.width(), draw_command.clipping_zone.height()) = (
    291.0,
    72.0,
    346.0,
    -2.6479187,
)
```

Which indicates that `egui` does send negative numbers.

If you'd like to, I can accept this PR as a temporary workaround, but we should change the check so that the rounded scissor params are strictly larger than 0.

And we definitely need to submit an issue to Egui.

PR #17, opened 2021-04-21
Fix an example

ZaMaZaN4iK:
Hi!

I've found that in ui.rs one import is missing - I've got compilation errors on my local machine when I've copy&pasted the example.

New import resolves the issue.


mvlabat:
Oh, indeed. And it compiles for this project, since `egui` here is one of the dependencies.. 
Nice catch, thank you!

PR #20, opened 2021-05-01
add `try_ctx_for_window` function

jakobhellermann:
This is necessary for bevy-inspector-egui to support multiple windows. Could you maybe release a small non-breaking version so that I can implement this without waiting for the next big release?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -253,6 +253,11 @@ impl EguiContext {
             .unwrap()
     }
 
+    /// Fallible variant of [`EguiContext::ctx_for_window`]. Make sure to set up the render graph by calling [`setup_pipeline`].
+    pub fn try_ctx_for_window(&self, window: WindowId) -> Option<&egui::CtxRef> {
+        self.ctx.get(&window)
+    }
+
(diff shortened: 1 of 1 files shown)


mvlabat:
What if we just change `ctx_for_window` to return an option? I'm personally fine with making a new minor release with this breaking change. I think it's easier to do so rather than making a patch release just to remove the `try` flavour later.

jakobhellermann:
In most cases `ctx_for_window` would be unwrapped if it returned an option. When creating a new window and doing  `.add_system_set(SystemSet::on_update(AppState::Done).with_system(ui_second_window.system()))` the UI system will only run when the window exists.

My use case is for adding a ui system to a Plugin which runs even if the window is not set up yet.

jakobhellermann:
I think having both methods is fine, because the panicking version will be used the most, and the `try_` version exists as a fallback for the more remaining cases. Same with `vec[idx]` vs `vec.get(idx)`.

mvlabat:
Sounds fair. And thanks for the PR, I'll release the new version today.

PR #21, opened 2021-05-01
fix regression due to lazy ctx initialization

### LINKED ISSUE #19: Egui crashes with "no entry found for key"
Today I tried updating both `bevy` and `bevy_egui` to their latest versions. Suddenly, in my UI system, I can no longer get the context for the primary window. The code looks like this:

```rust
pub fn ui(
    egui_ctx: ResMut<EguiContext>,
    // other stuff...
) {
    let ctx = egui_ctx.ctx(); // crashes
    
    // other stuff...
}
```

Oddly enough the example code seems to work just fine. I have no idea which interactions within the code might be causing the crash. I can upload the full code if it helps.

jakobhellermann:
should fix #19

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -228,8 +228,10 @@ pub struct EguiContext {
 
 impl EguiContext {
     fn new() -> Self {
+        let mut ctx = HashMap::default();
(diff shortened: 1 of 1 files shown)


jakobhellermann:
superseded by #24, see #19

Issue #22, opened 2021-05-02
Compile Errors with bevy_dylib on Windows10 - rust-lld: error: too many exported symbols (max 65535)


bayswaterpc:
I'm trying to get the bevy dynamic building feature working for faster build times.  I have been able to get the bevy crate's examples building with the dynamic feature , so I know it is possible for bevy to build with dynamic on my machine with the following these steps

https://github.com/bevyengine/bevy/issues/1126
https://bevyengine.org/learn/book/getting-started/setup/ - Fast Compile section

However, the bevy_egui is not working with those same steps.  It does work with the dynamic feature removed. I've created a fork showing the changes I've made, just adding the fast compile .carg/config.toml file & "dynamic" feature to bevy dependency.

https://github.com/bayswaterpc/bevy_egui

It gives the following error when building any of the example targets
[bevy_egui dynamic bevy error blob.txt](https://github.com/mvlabat/bevy_egui/files/6412088/bevy_egui.dynamic.bevy.error.blob.txt)

If anyone has been able to get this working on Windows would love some tips.

bayswaterpc:
Posted this a second too soon, did more digging saw the following advice on this thread and got it working
https://github.com/bevyengine/bevy/pull/2016

Will make a pull request for this

inodentry:
I am getting this error even on stable rust, which doesn't have the `-Zshare-generics` option. Hence, the advice given before for fixing the issue, does not actually fix the issue.

This is without even any plugins. I simply cloned the latest bevy main `dbd856de` and did:
```
cargo build --example load_gltf --features dynamic
```

It fails to link with the error:
```
  = note: rust-lld: error: too many exported symbols (max 65535)
```

PR #23, opened 2021-05-02
Settings required for dynamic builds on egui

bayswaterpc:
This adds required file and dependency adjustment for bevy/dynamic Fast Compile started with issue https://github.com/mvlabat/bevy_egui/issues/22 , 

Also adds example targets in the toml, 1 less thing for beginners to have to figure out.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -15,7 +15,7 @@ rustflags = ["-C", "link-arg=-fuse-ld=/usr/local/bin/zld", "-Zshare-generics=y"]
 
 [target.x86_64-pc-windows-msvc]
 linker = "rust-lld.exe"
-rustflags = ["-Zshare-generics=y"]
+rustflags = ["-Zshare-generics=n"]
 
 # Optional: Uncommenting the following improves compile times, but reduces the amount of debug info to 'line number tables only'
 # In most cases the gains are negligible, but if you are on macos and have slow compile times you should see significant gains.
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,23 @@
+
+# Add the contents of this file to `config.toml` to enable "fast build" configuration. Please read the notes below.
+
+# NOTE: For maximum performance, build using a nightly compiler
+# If you are using rust stable, remove the "-Zshare-generics=y" below.
+
+[target.x86_64-unknown-linux-gnu]
+linker = "/usr/bin/clang"
+rustflags = ["-Clink-arg=-fuse-ld=lld", "-Zshare-generics=y"]
+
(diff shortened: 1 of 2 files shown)


mvlabat:
Hi! Thanks for the PR.
I have a question: are you sure that these changes are needed to be able to compile an application with dynamic linking? I'm afraid that it'll force dynamic linking for all bevy_egui users, regardless of whether they want it or not. I believe it shouldn't be done it the library itself.

bayswaterpc:
Yes that would have, I meant to remove the "dyanmic"  in the dependencies & leave a commented out so it should have been.  

```
[dependencies]
bevy = { version = "0.5", default-features = false, features = ["render", "bevy_winit"] }
#use dynamic for  Fast Compile dependencies required , https://bevyengine.org/learn/book/getting-started/setup/
#bevy = { version = "0.5", default-features = false, features = ["bevy_wgpu", "x11", "png", "dynamic"] }
egui = "0.11.0"
webbrowser = { version = "0.5.5", optional = true }
winit = { version = "0.24.0", features = ["x11"], default-features = false }

[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
clipboard = { version = "0.5.0", optional = true }

[dev-dependencies]
version-sync = "0.9.2"

[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
bevy = { version = "0.5", default-features = false, features = ["bevy_wgpu", "x11", "png"] }
#use dynamic for  Fast Compile dependencies required , https://bevyengine.org/learn/book/getting-started/setup/
#bevy = { version = "0.5", default-features = false, features = ["bevy_wgpu", "x11", "png", "dynamic"] }
```

From my experience with the bug, the difference between building and not building was the inclusion of the "dynamic" dependency.  So  I don't think it will build dynamically with that removed.  However the config.toml could be causing some side effect which I'm not fully aware of.  I'll post on the bevy thread where I got my fix to check https://github.com/bevyengine/bevy/pull/2016.

mvlabat:
Sorry, I'm still convinced that this shouldn't be a part of the library. The faster build config also implies that every contributor has to have another linker installed. If they don't, they'll have a pretty confusing error in their console. All that can be documented in the README of course, but I believe that it's too much hassle for too little benefit. 

If a user wants to use faster builds for every project by default, they can still do that via environment variables. I see this config much more useful for applications, where developers spend most of their time while iterating on them.

I'll close this PR. But feel free to leave additional comments here if you have any. If more developers will show interest in this, I might reconsider it as well.

bayswaterpc:
That makes sense. My main goal was to spare someone else the pain of having
to figure out the windows dynamic build configuration on their project.
The pr and closed issue should pop up for those searching so mission
accomplished.  Thanks for your consideration and work on this!

PR #38, opened 2021-08-30
rustup egui to 0.14

CAD97:

PR #39, opened 2021-08-31
Bump egui to 0.14.2

kepet19:


CAD97:
#38 is nearly equivalent and prior

kepet19:
Ohh yeah sorry. I was not sure what to do, because I need minimum version 0.14.1 in my project, because of the bug https://github.com/emilk/egui/issues/632. 
I am all for the prior should go first, but the bug still exits in that version. that is why update it to 0.14.2. 
I am not trying to do duplicated work or upset any one. I am really sorry.

CAD97:
If egui sets `version = 0.Y.0` and you set `version = 0.Y.Z` in your project, you'll always get version 0.Y.A where A >= Z. Unless the bug fix impacts this library specifically, there's no reason for it to require it.

Additionally, even if you just use `version = 0.Y.0`, cargo will always choose to download version 0.Y.M for the maximal M unless you already have a lockfile or you otherwise specifically ask it not to.

kepet19:
Thanks that is nice to know! 
I am closing it now.

mvlabat:
I've just published 0.7.0. Thanks for the PR!

kepet19:
Thanks! that is nice!

Issue #47, opened 2021-12-24
Absorbing/suppressing keyboard/mouse input when hovering over egui window


rezural:
Hey

I'm having issues (and generally a nice time!) with using bevy-egui, and mouse/keyboard input still being consumed by the main bevy system code. This would be something nice to suppress somehow, probably opt-in, however I do feel that generally the happy path is for egui related input to be handled by egui, then suppressed (if possible).

Currently I am suppressing if egui.has_pointer_focus() (or somesuch), however, all my input processing code on the bevy side has to be made aware of egui, and it is getting unweildy.

I don't have enough of an understanding of bevy's input system, and where bevy-egui actually get's it's input from, to know where this is solvable from, so I'm just asking here as it seems the most obvious choice.

Is this solvable within bevy-egui? Is it currently do-able with bevy's input system?

Thanks

mvlabat:
Hi @rezural!

Thanks for the question. You're right, currently, the only way to do that is to pass Egui context to each system interested in the inputs and call `has_pointer_focus` (or the better choice might be `wants_keyboard_input` or `wants_pointer_input`). This is how I was doing it myself in my own project, but I agree - adding this to every system processing input can be cumbersome.

I think the only way `bevy_egui` can help is to introduce its own resources for input events, which would be emitted only if they are not consumed by Egui. This is something I'd probably explore closer to the 0.6 release, but I'm also happy to accept PRs if anyone wants to tackle this earlier.

rezural:
Ok, good to hear.

I will ask around in bevy if there are any ways to accomplish this using the event/input system. If I manage to get something working, I will send up a PR.

AlanRace:
It sounds like this issue is related to one I am currently having, so thought I would just add this here. Zooming in/out of a plot (holding `Ctrl` and using the mouse wheel) doesn't seem to work, however panning works fine (click and drag). Simple example below.

```rust
use bevy::prelude::*;
use bevy_egui::{
    egui::{
        self,
        plot::{Line, Plot, Value, Values},
    },
    EguiContext, EguiPlugin,
};

fn main() {
    App::build()
        .add_plugins(DefaultPlugins)
        .add_plugin(EguiPlugin)
        .add_system(ui_example.system())
        .run();
}

// Note the usage of `ResMut`. Even though `ctx` method doesn't require
// mutability, accessing the context from different threads will result
// into panic if you don't enable `egui/multi_threaded` feature.
[...]

PR #50, opened 2022-01-06
Drop events on focus

rezural:
This is a POC of dropping/suppressing of keyboard & mouse events, when the pointer is over any egui windows. Im not really that happy with it, but there is nothing wrong with it per-se, with the current state of bevy.

This is based off of jakobhellerman/bevy_egui#bevy-main, as it also depends on a slightly patched version of bevy:https://github.com/rezural/bevy/tree/input-derive-clone

I have gated it behind the drop_events_on_focus feature (feel free to rename this to something better), as while it is probably handy for bevy_egui consumers to opt into, it makes decisions about when to drop events, which are fairly arbitrary.

In reality, the application itself will probably want to decide which input state it is in (i.e. if there is a mouse pointer, or not), and a host of other concerns that will have to be fleshed out properly in bevy & the larger UI effort. I imagine this will come as a result of the bevy ui effort.


rezural:
The actual only relevant commit is 
https://github.com/mvlabat/bevy_egui/pull/50/commits/53c1773583f2dfc2a8f8bc2aa93de1c9c5fd05ec

rezural:
This shows an small example app that shows suppression of events, copy this into example/simple.rs,  run with:

`cargo run --example simple --features=drop_events_on_focus`


```rust

use bevy::prelude::*;
use bevy_egui::{egui, EguiContext, EguiPlugin};

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_plugin(EguiPlugin)
        .add_system(ui_example.system())
        .add_system(input_events.system())
        .run();
}

fn ui_example(egui_context: Res<EguiContext>) {
    egui::Window::new("Hello").show(egui_context.ctx(), |ui| {
        ui.label("world");
    });
    egui::Window::new("Hello2").show(egui_context.ctx(), |ui| {
        ui.label("world");
    });
}

fn input_events(mut ev_cursor: EventReader<CursorMoved>) {
    for event in ev_cursor.iter() {
        println!("Mouse move: {:?}", event);
    }
}

```

rezural:
Fixes #47

rezural:
One thing I only just thought of, is to leave most of the event consumption code the same in process_input system. And clear the input events in a system just after process_inputs.

If you think this PR has merit, I can rework this, so that most of process_input remains unchanged (except for the Input::clone())

ndavd:
Any updates on this?

vladbat00:
Thank you for this first POC! We've added this functionality in #369.

PR #53, opened 2022-01-09
run egui pass after `bevy_ui`

### LINKED ISSUE #52: bevy_egui race condition with bevy_ui
When mixing bevy_ui and bevy_egui in a project, it's random which one will be rendered on top of the other. 

Pic about egui being rendered below bevy ui (in this case there is a blueish background covering the whole screen): 
image
How it should look like:
unknown

It's random whether egui will be above bevyui for each start. 

Discord discussion about the issue in the Bevy Discord: https://discord.com/channels/691052431525675048/929761837363363950/929762436733624410

Dependencies:
bevy = "0.6.0"
bevy_egui = "0.10.0"

jakobhellermann:
should fix #52 

@Hilku can you confirm by putting
```toml
[patch.crates-io]
bevy_egui = { git = "https://github.com/jakobhellermann/bevy_egui", branch = "after-ui-pass" }
```

in your `Cargo.toml`?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -460,6 +460,8 @@ pub fn setup_pipeline(
             config.egui_pass,
         )
         .unwrap();
+
+    let _ = render_graph.add_node_edge("ui_pass_driver", config.egui_pass);
 }
 
 #[cfg(test)]


jakobhellermann:
This is a bit ugly because it uses a hardcoded string out of `bevy_ui` but I can't think of a better solution right now.

Hilku:
I tried it, and it seems to be better, but bevy_egui is still sometimes rendered "under" bevy_ui.

mvlabat:
I'll merge this until we can have a better fix. @Hilku unfortunately, I wasn't able to reproduce the problem when `bevy_egui` still gets rendered "under" `bevy_ui` sometimes. If you have an example, feel free to open a new issue and share it, I'll try to look into it.

PR #63, opened 2022-01-27
Add `multi_threaded` feature as passthrough to avoid needing separate egui dependency.

ndarilek:
… egui dependency.


mvlabat:
Fixed another occurrence in the docs and added a CI check.
Cheers!

PR #66, opened 2022-02-01
Remove link to megaui

lukors:
Megaui is deprecated and links back to this project.


mvlabat:
Good point. Thanks for the PR

PR #78, opened 2022-03-07
Update to egui 0.17

emilk:
## TODO:
* [x] Partial texture updates

Closes https://github.com/mvlabat/bevy_egui/issues/74
Closes https://github.com/mvlabat/bevy_egui/pull/75

I haven't yet figured out how to do a partial texture update (using https://docs.rs/wgpu/0.12.0/wgpu/struct.CommandEncoder.html#method.copy_buffer_to_texture), so instead I opted to store the full image, update part of it, and then recreate the full texture.

PR #79, opened 2022-03-08
Bevy 0.7

aevyrie:
This is based on #78 for egui 0.17. If that is merged, I will rebase on main.

This branch will be used to track bevy/main until the next release.

The biggest upgrade is support for low power mode in bevy (https://github.com/bevyengine/bevy/pull/3974). I've added a few lines to `process_output` that checks if egui has reported `needs_repaint`, and sends a bevy `RequestRedraw` event. I've also updated the examples to use the new `WinitSettings::desktop_app()` mode for reactive rendering.

This means bevy_egui apps can now use zero resources unless the user is providing input, or there is an egui animation in progress!

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,11 +19,16 @@ multi_threaded = ["egui/multi_threaded"]
 open_url = ["webbrowser"]
 
 [dependencies]
-bevy = { version = "0.6", default-features = false, features = [
+bevy = { git = "https://github.com/bevyengine/bevy", branch = "main", default-features = false, features = [
     "bevy_render",
     "bevy_winit",
-    "bevy_core_pipeline"
+    "bevy_core_pipeline",
 ] }
+# bevy = { version = "0.6", default-features = false, features = [
+#     "bevy_render",
+#     "bevy_winit",
+#     "bevy_core_pipeline"
+# ] }
 egui = { version = "0.17", features = ["convert_bytemuck"] }
 webbrowser = { version = "0.5.5", optional = true }
 winit = { version = "0.26.0", features = ["x11"], default-features = false }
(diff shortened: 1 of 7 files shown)


aevyrie:
@forbjok I merged your branch here to get the two window example working. This branch has some additional integrations to work with the new low power mode in 0.7.

Cargo.toml
```diff
@@ -19,11 +19,16 @@ multi_threaded = ["egui/multi_threaded"]
...
 
 [dependencies]
-bevy = { version = "0.6", default-features = false, features = [
+bevy = { version = "0.7", default-features = false, features = [
     "bevy_render",
     "bevy_winit",
-    "bevy_core_pipeline"
+    "bevy_core_pipeline",
 ] }
+# bevy = { version = "0.6", default-features = false, features = [
```
Therzok, line 27: Not a contributor, but this probably should be removed.

examples/simple.rs
```diff
@@ -1,9 +1,14 @@
-use bevy::prelude::*;
+use bevy::{prelude::*, window::PresentMode, winit::WinitSettings};
 use bevy_egui::{egui, EguiContext, EguiPlugin};
 
 fn main() {
     App::new()
         .add_plugins(DefaultPlugins)
+        .insert_resource(WinitSettings::desktop_app())
+        .insert_resource(WindowDescriptor {
+            present_mode: PresentMode::Mailbox,
```
mvlabat, line 9: Is it needed in this example?

examples/ui.rs
```diff
@@ -24,9 +24,15 @@ fn main() {
     App::new()
         .insert_resource(ClearColor(Color::rgb(0.0, 0.0, 0.0)))
         .insert_resource(Msaa { samples: 4 })
+        .insert_resource(WinitSettings::desktop_app())
         .init_resource::<UiState>()
         .add_plugins(DefaultPlugins)
+        .insert_resource(WinitSettings::desktop_app())
```
mvlabat, line 30: I guess one of these is redundant

mvlabat:
Thanks a lot for the PR! I'll make a new release soon.

aevyrie:
@mvlabat something I added that could use more discussion was the `present_mode` and the `WinitSettings` for low power. Maybe these should be removed from the examples, and instead added to a `dektop_app` example or something. It would be good to discuss how you can use these settings to improve input latency and power use.

PR #80, opened 2022-03-12
Add a funtion to get image id

Shatur:
I added this function because `add_image` requires `mut` and sometimes it inconvenient because you can't use it inside `show`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -347,7 +347,7 @@ impl EguiContext {
         id
     }
 
-    /// Returns associated Egui texture id
+    /// Returns associated Egui texture id.
     #[must_use]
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -346,6 +346,14 @@ impl EguiContext {
         log::debug!("Remove image (id: {:?}, handle: {:?})", id, image);
         id
     }
+
+    /// Returns associated Egui texture id
```
mvlabat, line 350:
```suggestion
    /// Returns associated Egui texture id.
```

mvlabat:
That's a good addition, thanks!

src/lib.rs
Shatur: @mvlabat done!

PR #81, opened 2022-03-12
Return egui::TextureId on removal

Shatur:
I wrapped the returned id into `egui::TextureId` since it usable only in this context. Just for consistency with other methods.


mvlabat:
Oh, how did I miss to update this method too?.. Good catch, thanks.
I'll have to merge it when making the 0.13 release, as it's a breaking change.

PR #95, opened 2022-04-24
Improved wgsl readability and minor optimisation

lain-dono:


mvlabat:
Nice work, thank you!

PR #96, opened 2022-04-25
new_tab support on open_url

Azorlogh:
This fixes hyperlinks opening in a new tab whether modifiers are pressed or not.

There is still an issue: opening a hyperlink in a new tab, then coming back to the previous tab, will leave the modifier keys pressed even if they've been released in the meantime. But I believe this is an issue with bevy and not bevy_egui.


mvlabat:
Perfect, thanks!

PR #98, opened 2022-04-26
Remove duplicate resource initiation EguiPipeline

lain-dono:


mvlabat:
Ha-ha, that's a nice catch, thanks.

PR #105, opened 2022-05-19
Window is open and checkbox is checked added to ui.rs example

tomuxmon:
Added demonstration how to use `egui::Window::open(self, &'open mut bool)`. Since I am a newcomer to rust I have been struggling with passing a mutable bool. Should be useful for other newcomers as well since closing or opening a window is such a common use case.

Hope that helps.

### DIFF
--- a/examples/ui.rs
+++ b/examples/ui.rs
@@ -31,15 +31,14 @@ fn main() {
             ..Default::default()
         })
         .init_resource::<UiState>()
-        .init_resource::<WindowState>()
         .add_plugins(DefaultPlugins)
         .add_plugin(EguiPlugin)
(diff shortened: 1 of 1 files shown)


tomuxmon:
Yes, it can. But then the default state will be a closed window unless we provide a manual Default implementation for `UiState`. Is there a way to make `is_open` default to true and leave `#[derive(Default)]` on `UiState`?

tomuxmon:
Using only `UiState` now. Left default as is but introduced `configure_ui_state` startup function to set `is_window_open` to `true`.

tomuxmon:
can we merge it or is there something else to be fixed?

mvlabat:
Merged. Thank you!

PR #110, opened 2022-07-02
Add feature to disable egui builtin fonts

### LINKED ISSUE #107: Add feature to disable egui builtin fonts
I would like to disable the builtin egui fonts to save space.
egui includes a feature in its default feature set that includes them.
bevy_egui should mirror that with a default feature that pulls in that egui feature that can be disabled.
```toml
[features]
default = ["manage_clipboard", "open_url", "default_fonts"]
immutable_ctx = []
manage_clipboard = ["arboard", "thread_local"]
open_url = ["webbrowser"]
default_fonts = ["egui/default_fonts"]

[dependencies]
bevy = { version = "0.7", default-features = false, features = ["bevy_render", "bevy_core_pipeline"] }
egui = { version = "0.18", default-features = false, features = ["bytemuck"] }
```

iTitus:
Closes #107

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,14 +14,15 @@ all-features = true
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 [features]
-default = ["manage_clipboard", "open_url"]
+default = ["manage_clipboard", "open_url", "default_fonts"]
 immutable_ctx = []
 manage_clipboard = ["arboard", "thread_local"]
 open_url = ["webbrowser"]
+default_fonts = ["egui/default_fonts"]
 
 [dependencies]
 bevy = { version = "0.7", default-features = false, features = ["bevy_render", "bevy_core_pipeline"] }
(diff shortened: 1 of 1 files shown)


mvlabat:
Hey! Thanks for the PR. Do you think it makes sense to wrap the rest of the features the same way? I'm not really sure how libraries usually approach that. Users can as well add needed features by specifying `egui` as a direct dependency and adding them there. But I don't know whether it's a good UX.

iTitus:
This is only for features in egui that one wants to disable.
By disabling all default features and putting them behind our default features we give the user a way to choose.

One can always add features, but removing is impossible.

mvlabat:
Yeah, with that I absolutely agree. bevy_egui users should have a way to run `egui` without default features.
I was wondering specifically about non-default features, whether it's a common practice to wrap them as well.

As for this feature, I'll merge it as soon as I start preparing the new release for Bevy 0.8.

iTitus:
I think non-default features only get wrapped if the package makes explicit use of them.

PR #111, opened 2022-07-02
Update to work with bevy main (and eventually 0.8)

DGriffin91:
This needs a more thorough review. But it looks like the examples are working, and it's working in at least one of my projects.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ manage_clipboard = ["arboard", "thread_local"]
 open_url = ["webbrowser"]
 
 [dependencies]
-bevy = { version = "0.7", default-features = false, features = ["bevy_render", "bevy_core_pipeline"] }
+bevy = { git = "https://github.com/bevyengine/bevy", default-features = false, features = ["bevy_render", "bevy_core_pipeline"] }
 egui = { version = "0.18", features = ["bytemuck"] }
 webbrowser = { version = "0.7", optional = true }
 
@@ -31,7 +31,7 @@ thread_local = { version = "1.1.0", optional = true }
 [dev-dependencies]
 once_cell = "1.9.0"
 version-sync = "0.9.2"
-bevy = { version = "0.7", default-features = false, features = [
+bevy = { git = "https://github.com/bevyengine/bevy", default-features = false, features = [
     "x11",
     "png",
     "bevy_pbr",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ manage_clipboard = ["arboard", "thread_local"]
 open_url = ["webbrowser"]
 
 [dependencies]
-bevy = { git = "https://github.com/bevyengine/bevy", default-features = false, features = ["bevy_render", "bevy_core_pipeline", "bevy_asset"] }
+bevy = { version = "0.8", default-features = false, features = ["bevy_render", "bevy_core_pipeline", "bevy_asset"] }
 egui = { version = "0.18", features = ["bytemuck"] }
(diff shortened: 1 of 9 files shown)


zimond:
This is broken as [bevy#4402](https://github.com/bevyengine/bevy/pull/4402) landed

DGriffin91:
Just messed with it a bit and it seems like this will require a bit of a more significant re-work now as the extract systems currently require mutable access to resources in the main world.

zimond:
Yes and use immutable `Res` with clone could make the examples run, but with only a black screen [code](https://github.com/zimond/bevy_egui/commit/65d8cd9e22dc534612e3436d09388b362c3addbe) I didn't have the time to dig in

DGriffin91:
@[zimond](https://github.com/zimond) I tried using the clone with `egui_render_output` in the extract instead of take and it looks like it's working for me. I would assume the clone here is less efficient than take, but I'm not sure if it's enough slower to matter.

zimond:
I think some clones won't hurt the perf that much

mvlabat:
Hi! Thank you for the PR! I think cloning is fine for now. Maybe, if we see an issue with performance, we might want to wrap it into a `Mutex`, if copying a lot of bytes around will become more expensive than dealing with locking.
I've also pushed a small commit to refactor the example, to make the error of initialization more visible (it panic now if we forget to initialize the camera properly, instead of making it silent) and fix the linters.

zimond:
Closing a window on this branch emits panic:

```rust
let extracted_window = &world.get_resource::<ExtractedWindows>().unwrap().windows[&self.window_id];
```

```bash
thread 'main' panicked at 'no entry found for key' egui_node.rs:328:14
```

DGriffin91:
Hmm, I tried this
```rust
        let extracted_windows = &world.get_resource::<ExtractedWindows>().unwrap().windows;

        let extracted_window =
            if let Some(extracted_window) = extracted_windows.get(&self.window_id) {
                extracted_window
            } else {
                return Ok(());
            };
 ```
But now when I close the window the app continues to run with the window closed and says:
```
2022-07-18T16:59:55.466789Z  INFO bevy_winit: Skipped event for closed window: WindowId(00000000-0000-0000-0000-000000000000)
2022-07-18T16:59:55.467187Z  INFO bevy_winit: Skipped event for closed window: WindowId(00000000-0000-0000-0000-000000000000)
2022-07-18T16:59:55.470299Z  INFO bevy_winit: Skipped event for closed window: WindowId(00000000-0000-0000-0000-000000000000)
```

mvlabat:
Any chance that this is an issue on the Bevy's end? Not really sure what can cause this in `bevy_egui` itself. 
I haven't got a chance to look into it myself, but does it happen with every example? Or is it specific to the multiple windows example?

If it happens in every example, I'd suggest to try to reproduce it in a clean Bevy project. And if it reproduces there without `bevy_egui`, we need to file an issue in their repo.

DGriffin91:
@mvlabat It appears that it is indeed an issue with bevy: https://github.com/bevyengine/bevy/issues/5384

CGMossa:
Is this issue blocking this PR or has it been fixed?

mvlabat:
Definitely not a blocker for us. I don't know if it has been fixed in Bevy's main, but we'll merge this PR regardless once 0.8 is released

philpax:
Bevy 0.8's now out - I've just built this fork with it (opened a PR for it here https://github.com/DGriffin91/bevy_egui/pull/2) and everything seems to work fine 👍

mvlabat:
Thank you for your help with the update! I'll release it soon

PR #116, opened 2022-08-01
Bump bevy to 0.8 in readme

agluszak:


mvlabat:
Ah, good catch. Thanks :)

Issue #61, opened 2022-01-24
Mouse events are ignored after cursor leaves window


jakobhellermann:
Before https://github.com/mvlabat/bevy_egui/commit/a496e4984b7ab9224d2a33571187fd188da03add it was possible to start dragging somewhere in the window (for example a slider), then move the cursor out of the window while still having the mouse button pressed and then returning.

After the change, the mouse events will stop firing as soon as the cursor leaves the window.

Removing the `if` in

https://github.com/mvlabat/bevy_egui/commit/a496e4984b7ab9224d2a33571187fd188da03add#diff-420f4e08a25299e0e1fb09f4757e1d5d027a2278f37030c09ab9c06786cfa52eR155-R157

fixes this, although I'm not sure why, because the `if` condition is only `false` for one frame.

Before:

https://user-images.githubusercontent.com/22177966/150784489-6eb90928-5989-45bf-a7f9-f7e13efaa064.mp4

After:

https://user-images.githubusercontent.com/22177966/150785233-1d3dc145-e37e-424f-a990-8d8055570ef5.mp4

mvlabat:
I believe it's by Egui's design that when you emit `egui::Event::PointerGone` and don't push another `egui::Event::PointerMoved` event during the same frame, it forgets that the mouse button is still being pressed and the cursor started its movement from a certain widget. I can reproduce the same behaviour in the [Egui's demo](https://emilk.github.io/egui/). So I'd even consider the previous behaviour to be a bug, as pushing `egui::Event::PointerMoved` right after `egui::Event::PointerGone` doesn't quite make sense.

Though I agree that being able to keep dragging a slider outside a window is a good thing. But I guess it's an issue of Egui itself.

coderedart:
I am kinda stuck on this. 
1.  so, with eframe, if i click and drag beyond the bounds of native window, and release the mouse button, egui will stop the drag. it works as expected.
2. but, with bevy_egui, if i click and drag beyond the bounds of native window, even after releasing the mouse button, egui will still continue the drag, which leads to egui not using the events at all. so, I have to click somewhere again for it to stop the drag.

https://user-images.githubusercontent.com/24411704/166121391-cf3791c9-e472-463f-aa41-66372a5a7a85.mp4

[...]

PR #121, opened 2022-09-09
Correctly handle releasing mouse buttons when not hovering a window

TheRawMeatball:


src/systems.rs
```diff
@@ -128,8 +128,38 @@ pub fn process_input(
...
+                        events.push(egui::Event::PointerButton {
+                            pressed: false,
+                            button,
+                            modifiers: Default::default(),
+                            pos: Default::default(),
+                        });
+                    }
+                }
+                events.push(egui::Event::PointerGone);
+            }
```
mvlabat, line 161: Hi! Thanks for the PR.

Is the goal here to fix the behaviour for any platform? I believe this branch will be entered only for Safari, as this is the only browser that emits `CursorLeft` and `CursorEntered` on mouse button release.

src/systems.rs
```diff
@@ -128,8 +128,38 @@ pub fn process_input(
     // When a user releases a mouse button, Safari emits both `CursorLeft` and `CursorEntered`
     // events during the same frame. We don't want to reset mouse position in such a case, otherwise
     // we won't be able to process the mouse button event.
-    if cursor_left_window.is_some() && cursor_left_window != cursor_entered_window {
-        egui_context.mouse_position = None;
+    if let Some(window_id) = &cursor_left_window {
+        if cursor_left_window != cursor_entered_window {
```
mvlabat, line 132: Hi! Thanks for the PR.

Is the goal here to fix the behaviour for any platform? I believe this branch will be entered only for Safari, as this is the only browser that emits `CursorLeft` and `CursorEntered` on mouse button release.

TheRawMeatball: That seems off since this change fixed unexpected behavior on desktop linux.

TheRawMeatball: Furthermore, this branch needs to be possible to entered on all platforms - after all it's the only branch where `mouse_position` is reset.

mvlabat: Hm.. I'm not sure it's working as intended. If I keep dragging the mouse pointer and it leaves the window, it'll fire the release button event immediately. Yet in the `main` version we keep reading mouse events when outside of a window, which allows us to continue dragging sliders for example (see #61).

https://user-images.githubusercontent.com/2943388/190848814-f1d668ab-62d0-4d2b-a88c-514d307b2a3a.mov

mvlabat: I'm also not sure I understand what issue the PR is trying to address (or whether the issue indeed exists).

During this recording (the `main` branch), I dragged the cursor outside of the window, released it when being outside, moved the cursor back to the window, and we can see that `bevy_egui` didn't continue drawing to the canvas. If I understand correctly, your point is that we don't fire mouse release events when releasing the cursor outside of the window. But if that was the case, wouldn't it keep drawing when I moved the cursor back to the window?

https://user-images.githubusercontent.com/2943388/190848987-c19e275d-8942-410c-9aa2-4aa1ec5595a0.mov

TheRawMeatball: Yes, my main issue is that when the mouse is released outside the window, the release event is never fired, and in a closed-source project i work on it does cause the mean that drawing continues.

TheRawMeatball: And I just tested the main branch - the drawing continues when i release the button outside the window and move the cursor back in.

TheRawMeatball: [egui-show-bug-.webm](https://user-images.githubusercontent.com/6781733/190850904-b2e274dc-7744-4ba4-ba4d-dbee04e16050.webm)

mvlabat: Oh, I see, thanks for clarifying. I was indeed able to reproduce it on my KDE laptop.

Turns out, for some platforms, `winit` fires `CursorLeft` together with button input events when releasing a button outside of a window. I've just pushed  #123 with the fix that also preserves the dragging behaviour. Could you please test it and let me know if it works for you?

TheRawMeatball: It seems to work :)

mvlabat: Merged and released. Thanks again for flagging it!

PR #127, opened 2022-10-07
update to bevy 0.9 dev

terhechte:
One of my projects has a rendering bug that is fixed on Bevy/main. Due to breaking changes in `0.9 dev`, I couldn't use bevy_egui anymore. This PR adopts all the bevy changes required to run bevy_egui with the current state of bevy 0.9 dev. Not sure if you already want to merge it, but I wanted to share it already in case anybody else is in my situation.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,18 +21,18 @@ open_url = ["webbrowser"]
 default_fonts = ["egui/default_fonts"]
 
 [dependencies]
-bevy = { git = "https://github.com/bevyengine/bevy", branch = "main", default-features = false, features = ["bevy_render", "bevy_core_pipeline", "bevy_asset"] }
+bevy = { version = "0.9.0", default-features = false, features = ["bevy_render", "bevy_core_pipeline", "bevy_asset"] }
 egui = { version = "0.19.0", default-features = false, features = ["bytemuck"] }
-webbrowser = { version = "0.7", optional = true }
+webbrowser = { version = "0.8.2", optional = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 arboard = { version = "2.0.1", optional = true }
 thread_local = { version = "1.1.0", optional = true }
 
 [dev-dependencies]
-once_cell = "1.9.0"
-version-sync = "0.9.2"
-bevy = { git = "https://github.com/bevyengine/bevy", branch = "main", default-features = false, features = [
+once_cell = "1.16.0"
+version-sync = "0.9.4"
+bevy = { version = "0.9.0", default-features = false, features = [
     "x11",
     "png",
     "bevy_pbr",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ open_url = ["webbrowser"]
 default_fonts = ["egui/default_fonts"]
 
 [dependencies]
-bevy = { version = "0.8", default-features = false, features = ["bevy_render", "bevy_core_pipeline", "bevy_asset"] }
+bevy = { git = "https://github.com/bevyengine/bevy", branch = "main", default-features = false, features = ["bevy_render", "bevy_core_pipeline", "bevy_asset"] }
 egui = { version = "0.19.0", default-features = false, features = ["bytemuck"] }
 webbrowser = { version = "0.7", optional = true }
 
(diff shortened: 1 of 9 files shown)


src/lib.rs
```diff
@@ -116,10 +114,58 @@ impl Default for EguiSettings {
...
+#[derive(Resource)]
+pub struct EguiRenderOutputContainer(pub HashMap<WindowId, EguiRenderOutput>);
+
+impl Deref for EguiRenderOutputContainer {
+    type Target = HashMap<WindowId, EguiRenderOutput>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
```
MrGVSV, line 127: You can actually just use Bevy's `Deref` and `DerefMut` derives to achieve this.

src/lib.rs
```diff
@@ -433,10 +479,19 @@ impl Plugin for EguiPlugin {
     fn build(&self, app: &mut App) {
         let world = &mut app.world;
         world.insert_resource(EguiSettings::default());
-        world.insert_resource(HashMap::<WindowId, EguiInput>::default());
-        world.insert_resource(HashMap::<WindowId, EguiOutput>::default());
-        world.insert_resource(HashMap::<WindowId, WindowSize>::default());
-        world.insert_resource(HashMap::<WindowId, EguiRenderOutput>::default());
+        world.insert_resource(EguiRenderInputContainer(
+            HashMap::<WindowId, EguiInput>::default(),
+        ));
```
MrGVSV, line 484: I would just make all these new resources derive `Default`. Then you can just do `EguiRenderInputContainer::default()` (or even `world.init_resource::<EguiRenderInputContainer>()`).

src/systems.rs
```diff
@@ -229,6 +233,7 @@ pub fn process_input(
             if !event.char.is_control() {
                 input_resources
                     .egui_input
+                    .0
```
MrGVSV, line 236: These `.0`'s won't be needed if you derive `DerefMut`.

src/lib.rs
mvlabat: And will also make it possible to initialize resources if some users decide not to add the egui plugin, by adding the resources and systems manually (with an intention to replace some of them).

Cargo.toml
```diff
@@ -21,7 +21,7 @@ open_url = ["webbrowser"]
 default_fonts = ["egui/default_fonts"]
 
 [dependencies]
-bevy = { version = "0.8", default-features = false, features = ["bevy_render", "bevy_core_pipeline", "bevy_asset"] }
+bevy = { git = "https://github.com/bevyengine/bevy", branch = "main", default-features = false, features = ["bevy_render", "bevy_core_pipeline", "bevy_asset"] }
```
kerkmann, line 24:
```suggestion
bevy = { version = "0.9", default-features = false, features = ["bevy_render", "bevy_core_pipeline", "bevy_asset"] }
```

Bevy 0.9 is now released. :)

Cargo.toml
```diff
@@ -32,7 +32,7 @@ thread_local = { version = "1.1.0", optional = true }
 [dev-dependencies]
 once_cell = "1.9.0"
 version-sync = "0.9.2"
-bevy = { version = "0.8", default-features = false, features = [
+bevy = { git = "https://github.com/bevyengine/bevy", branch = "main", default-features = false, features = [
```
kerkmann, line 35:
```suggestion
bevy = { version = "0.9", default-features = false, features = [
```

Bevy 0.9 is now released. :)

kerkmann (changes requested):
Bevy 0.9 is now released. :)

mvlabat:
Will release the update tomorrow (it's midnight here in Ukraine 🙂).

mvlabat:
Merging the PRs, thanks!
I'm also going to address @MrGVSV's comments and do some cleaning up in a separate one.

Cargo.toml
```diff
@@ -21,18 +21,18 @@ open_url = ["webbrowser"]
 default_fonts = ["egui/default_fonts"]
 
 [dependencies]
-bevy = { version = "0.8", default-features = false, features = ["bevy_render", "bevy_core_pipeline", "bevy_asset"] }
+bevy = { version = "0.9.0", default-features = false, features = ["bevy_render", "bevy_core_pipeline", "bevy_asset"] }
```
kerkmann, line 24: @mvlabat 
Thanks for merging it! <3

btw; Is there a reason why the version is fixed to `0.9.0` instead of `0.9`? :)

mvlabat: Hi! No particular reason. In `0.9.0` behaves the same way as `^0.9.0` (and `0.9`). If `0.9.1` is released, this won't prevent to install it.
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

kerkmann: Thanks For Letting Me Know. :3

PR #128, opened 2022-10-10
Fix window resize cursor icon on Windows

### LINKED ISSUE #126: Cannot resize main window using mouse
Thanks for the very useful crate that `bevy_egui` is.

Trying to resize window using mouse doesn't work, as confirmed using example `simple.rs`.
When putting mouse cursor near a window edge, the cursor doesn't change into resizing style (thin double arrow cursor) and there is no way to resize the window.

Versions (current): 
bevy = "0.8.1"
bevy_egui = "0.16.1"

Without `bevy_egui` this simple example works:
```
use bevy::prelude::*;

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .run();
}

```

When adding `bevy_egui` resizing window with mouse doesn't work anymore:
```
use bevy::prelude::*;
use bevy_egui::EguiPlugin;

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_plugin(EguiPlugin)
        .run();
}

```

By default, for bevy `WinitSettings` the value for `focused_ mode` is `UpdateMode::Continuous` which is needed for games, and we can observe the problem that the cursor doesn't change to resizing style when near window edges.

And when we change `focused_mode` to `UpdateMode::Reactive`, with adding:
 ```
.insert_resource(WinitSettings {
    focused_mode: UpdateMode::Reactive {
        max_wait: std::time::Duration::from_secs(1),
    },
    ..Default::default()
})
```
... the cursor at first correctly switches to resize style when nearing window edge, but then reverts to arrow within less than a second, which matches the 1 second wake up delay set with `UpdateMode::Reactive` above.

So apparently the resize action is cancelled when bevy wakes up, or more probably at each beginning of its cycle (as witnessed with `UpdateMode::Continuous`). I couldn't find where exactly though.

chronicl:
This fixes a Windows specific problem:
Currently windows created by bevy with bevy_egui installed show the default cursor icon when hovering over the edges of the window, although a resize cursor icon for window resizing should be shown instead. 

This was implemented in a similar fashion to how `egui-winit` handles this problem, see [here](https://github.com/emilk/egui/blob/367378d75dcc1670418ce464cb1aafa4c8b2387b/crates/egui-winit/src/lib.rs#L614).

Probably partially fixes #126.

### DIFF
--- a/src/systems.rs
+++ b/src/systems.rs
@@ -349,6 +349,7 @@ pub fn process_output(
     #[cfg(feature = "manage_clipboard")] mut egui_clipboard: ResMut<crate::EguiClipboard>,
     mut windows: Option<ResMut<Windows>>,
     mut event: EventWriter<RequestRedraw>,
+    #[cfg(windows)] mut last_cursor_icon: Local<HashMap<WindowId, egui::CursorIcon>>,
 ) {
     for (window_id, ctx) in egui_context.ctx.iter_mut() {
         let full_output = ctx.end_frame();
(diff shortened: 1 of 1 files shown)


mvlabat:
Seems like it does fix the issue! Thank you for the PR.

PR #131, opened 2022-11-09
add pub use bevy 

kotlin2018:


mvlabat:
Hi! Thanks for the PR.
I don't think that's something many plugin users will find useful. I can't really imagine the case when people would want to import and use `bevy_egui::bevy` instead of `bevy` directly.
I'll close this. But if you have any arguments on why this might be needed, feel free to post them here.

PR #138, opened 2022-12-08
Default the context instead of panicking

Aceeri:
This ends up being somewhat of an issue on some platforms now, since the primary window *might* not be up before the startup system runs.


mvlabat:
Hi! Thanks for the PR.
Could you give an example of a platform where this is an issue, and the steps to reproduce it?

I'm also curious if it's an option to use `try_ctx_for_window_mut` in your case.

If we are to go this route (where we always create a default context), maybe we can drop the `try_` methods then. But I'm not 100% sold on it yet..

Aceeri:
I think this ended up being a misleading end for an issue I was debugging, going to close this out for now.

PR #139, opened 2022-12-09
Update to support egui 0.20

no-materials:


mvlabat:
Thank you very much for the PR! Will release the new version soon.

PR #143, opened 2023-01-05
Bump `arboard` version to reduce duplicate deps

alice-i-cecile:
Should fix https://github.com/Leafwing-Studios/Emergence/issues/194.

This is a first attempt: we'll see if CI passes and fix up any breaking changes if needed.


TimJentzsch:
With the syntax error fixed, `cargo check --all-features` and `cargo test --all-features` both pass for me :)

TimJentzsch (approved):

PR #147, opened 2023-01-29
Implement better multi-window support

TheRawMeatball:
Make it so all windows support rendering egui output by default.


mvlabat:
Hi! Thanks a lot for the PR. I'm merging it but will release a bit later, I hope to include it in the Bevy 0.10 upgrade.

PR #148, opened 2023-02-07
Update for bevy 0.10

DGriffin91:
Having a go at trying to update this for current bevy main. Tried to use the window Entity as the window id, but running into the issue that it is currently designed expecting that the id for the primary window will already exist before the window does.

Most relevant PRs to look at:
[Windows as Entities #5589](https://github.com/bevyengine/bevy/pull/5589)
[Migrate engine to Schedule v3 #7267](https://github.com/bevyengine/bevy/pull/7267)

### DIFF
--- a/examples/render_to_image_widget.rs
+++ b/examples/render_to_image_widget.rs
@@ -58,6 +58,7 @@ fn setup(
             usage: TextureUsages::TEXTURE_BINDING
                 | TextureUsages::COPY_DST
                 | TextureUsages::RENDER_ATTACHMENT,
+            view_formats: &[],
         },
         ..default()
     };
@@ -107,7 +108,7 @@ fn setup(
             },
             camera: Camera {
                 // render before the "main pass" camera
-                priority: -1,
+                order: -1,
                 target: RenderTarget::Image(image_handle),
                 ..default()
             },
@@ -150,12 +151,13 @@ fn render_to_image_example_system(
     preview_cube_query: Query<&Handle<StandardMaterial>, With<PreviewPassCube>>,
     main_cube_query: Query<&Handle<StandardMaterial>, With<MainPassCube>>,
     mut materials: ResMut<Assets<StandardMaterial>>,
+    windows: Query<Entity, With<Window>>,
 ) {
     let cube_preview_texture_id = egui_ctx.image_id(&cube_preview_image).unwrap();
     let preview_material_handle = preview_cube_query.single();
     let preview_material = materials.get_mut(preview_material_handle).unwrap();
 
-    let ctx = egui_ctx.ctx_mut();
+    let ctx = egui_ctx.ctx_for_window_mut(windows.iter().next().unwrap());
     let mut apply = false;
     egui::Window::new("Cube material preview").show(ctx, |ui| {
         ui.image(cube_preview_texture_id, [300.0, 300.0]);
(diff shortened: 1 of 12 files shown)


DGriffin91:
Initially got the examples working with some string and bubblegum. Wondering if the instances resources with `HashMap<Entity, Something>` should now be components added to the window?

src/egui_node.rs
```diff
@@ -171,9 +175,9 @@ pub struct EguiNode {
 
 impl EguiNode {
     /// Constructs Egui render node.
-    pub fn new(window_id: WindowId) -> Self {
+    pub fn new(window_entity: Entity) -> Self {
```
Vrixyz, line 178: I tried to update to bevy main before seeing your PR ; and I did a `type WindowEntity = Entity`, to help me understand where the various `Entity` references were supposed to reference a window.

Would that be interesting ? It might be discussed in another issue, upgrading can already be a big enough scope.

DGriffin91: That could potentially help make things a bit more clear, though I wonder if we should really be storing the entity all over the place vs turning most things into components on the window entity.

mvlabat: I'd like us to try the components approach. Not sure what that would look like exactly, but since windows are entities now, storing additional data as components sounds more idiomatic

examples/two_windows.rs
```diff
@@ -24,40 +24,49 @@ fn main() {
...
-    let mut graph = render_app.world.get_resource_mut::<RenderGraph>().unwrap();
 
-    bevy_egui::setup_pipeline(
-        &mut graph,
-        bevy_egui::RenderGraphConfig {
-            window_id: *SECOND_WINDOW_ID,
-            egui_pass: SECONDARY_EGUI_PASS,
-        },
-    );
+    render_app.add_system_to_schedule(ExtractSchedule, init_second_window);
```
johanhelsing, line 28: this method no longer exists in bevy main. see: https://github.com/DGriffin91/bevy_egui/pull/4

johanhelsing: Oops I replied in the wrong place, was meaning to add it  the line in lib.rs. regardless, that linked PR makes it build with latest main.

examples/render_to_image_widget.rs
```diff
@@ -150,12 +151,13 @@ fn render_to_image_example_system(
...
     main_cube_query: Query<&Handle<StandardMaterial>, With<MainPassCube>>,
     mut materials: ResMut<Assets<StandardMaterial>>,
+    windows: Query<Entity, With<Window>>,
 ) {
     let cube_preview_texture_id = egui_ctx.image_id(&cube_preview_image).unwrap();
     let preview_material_handle = preview_cube_query.single();
     let preview_material = materials.get_mut(preview_material_handle).unwrap();
 
-    let ctx = egui_ctx.ctx_mut();
+    let ctx = egui_ctx.ctx_for_window_mut(windows.iter().next().unwrap());
```
Vrixyz, line 160: that's a loss of ergonomics indeed, I look forward to what can be done component-wise :) ; egui_context stored in a component attached to a window is my first not-much-informed thought

DGriffin91: Now this looks like:
```rs
egui_ctx: Query<&EguiContext, With<Window>>,
[...]
let ctx = egui_ctx.iter().next().unwrap();
```
Thoughts?
The `With<Window>` isn't really needed, but maybe clarifies things. idk if it makes sense or not.

mvlabat:
Thanks a lot for the help! I merged the update as part of #159, so closing this one.

PR #155, opened 2023-02-24
pass raw time to bevy_egui

jakobhellermann:
Time in egui is mostly used for animation. These animations should run regardless of whether `Time` is currently paused.


mvlabat:
Thank you! I'll merge the PR after the CI checks and include it in the release.

PR #159, opened 2023-03-06
Bevy 0.10

mvlabat:
Based on #148

### DIFF
--- a/examples/simple.rs
+++ b/examples/simple.rs
@@ -11,8 +11,11 @@ fn main() {
         .run();
 }
 
-fn ui_example_system(mut egui_context: ResMut<EguiContext>) {
-    egui::Window::new("Hello").show(egui_context.ctx_mut(), |ui| {
-        ui.label("world");
-    });
+fn ui_example_system(mut egui_context: ResMut<EguiContext>, windows: Query<Entity, With<Window>>) {
+    egui::Window::new("Hello").show(
+        egui_context.ctx_for_window_mut(windows.iter().next().unwrap()),
+        |ui| {
+            ui.label("world");
+        },
+    );
 }
--- a/examples/simple.rs
+++ b/examples/simple.rs
@@ -7,11 +7,11 @@ fn main() {
         .add_plugin(EguiPlugin)
         // Systems that create Egui widgets should be run during the `CoreStage::Update` stage,
         // or after the `EguiSystem::BeginFrame` system (which belongs to the `CoreStage::PreUpdate` stage).
-        .add_system(ui_example)
+        .add_system(ui_example_system)
         .run();
 }
 
-fn ui_example(mut egui_context: ResMut<EguiContext>) {
+fn ui_example_system(mut egui_context: ResMut<EguiContext>) {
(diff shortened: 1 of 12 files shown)


mvlabat:
I'm seeing the following warnings in the `ui` example:

> epaint: WARNING: pixels_per_point (dpi scale) have changed between text layout and tessellation. You must recreate your text shapes if pixels_per_point changes.

I'd be grateful if anyone could pick up this issue. I won't have time to return to the PR until the evening.

DGriffin91:
I've been trying to track down the `pixels_per_point` thing. Haven't figured it out yet. This is the commit that added the warning:
https://github.com/emilk/egui/commit/39b63f6163c7aabbf20138b505f0c86d62977408

In the ui example, it comes from changing the `egui_settings.scale_factor` in `update_ui_scale_factor_system()`

We then set `egui_input.pixels_per_point` in `update_window_contexts()` which is called from `process_input_system()`

We get the shapes in `process_output_system()`
```rs
let full_output = ctx.end_frame();
let egui::FullOutput {
    platform_output,
    shapes,
    textures_delta,
    repaint_after,
} = full_output;

egui_render_output.shapes = shapes;
```

I think it only shows the warning for one frame if `scale_factor` changes.

examples/simple.rs
```diff
@@ -1,18 +1,18 @@
...
+        // Systems that create Egui widgets should be run during the `CoreSet::Update` set,
+        // or after the `EguiSet::BeginFrame` system (which belongs to the `CoreSet::PreUpdate` set).
         .add_system(ui_example_system)
         .run();
 }
 
-fn ui_example_system(mut egui_context: ResMut<EguiContext>) {
-    egui::Window::new("Hello").show(egui_context.ctx_mut(), |ui| {
+fn ui_example_system(mut egui_ctx: Query<&mut EguiContext, With<PrimaryWindow>>) {
+    egui::Window::new("Hello").show(egui_ctx.single_mut().get_mut(), |ui| {
```
Vrixyz, line 15: can it be simplified like this?

```suggestion
fn ui_example_system(egui_ctx: Query<&EguiContext, With<PrimaryWindow>>) {
    egui::Window::new("Hello").show(egui_ctx.single(), |ui| {
```

mvlabat: Only if we implement `Deref` and `DerefMut` for `EguiContext`, giving people an interface to borrow the context immutably, which I'm hesitant to do due to the following:
https://github.com/mvlabat/bevy_egui/blob/7151c40f8564650fdfad5eaa15e72d055d0a5893/src/lib.rs#L247-L256

mvlabat:
@DGriffin91 thanks for looking into this! That definitely saved me time investigating. It seems that I was able to fix the warning here: 7151c40.

mvlabat:
The PR is almost finished. I'll let people review the API changes, and tomorrow morning I'll push the last bits of the documentation fixes and make a release.

examples/simple.rs
mvlabat: I ended up introducing the `EguiContexts` system param, to combine the best of both worlds: low boilerplate and mutable borrowing by default.

PR #162, opened 2023-03-10
Fix recreation of EguiContext on startup

### LINKED ISSUE #160: FontDefinitions being overwritten
I've been using the following code in a startup system to set custom fonts:

```rust
const TITLE_FONT_NAME: &str = "MY_FONT_NAME";
let mut fonts = FontDefinitions::default();
fonts.font_data.insert(
    TITLE_FONT_NAME.into(),
    FontData::from_static(include_bytes!(
        "path/to/font.ttf"
    )),
);
fonts
    .families
    .entry(FontFamily::Name(TITLE_FONT_NAME.into()))
    .or_default()
    .push(TITLE_FONT_NAME.into());
egui_context.ctx_mut().set_fonts(fonts);
```

I used this by setting
```rust
FontId {
  size: 30.0,
  family: FontFamily::Name(MY_TITLE_FONT.into()),
}
```
as the font property of a `RichText` object displayed in the UI.

This worked fine in bevy 0.9 / bevy_egui 0.19. However, since migrating to bevy 0.10 / bevy_egui 0.20, this doesn't work anymore and it panics when I try to use the font, saying

```
thread 'Compute Task Pool (18)' panicked at 'FontFamily::Name("MY_FONT_NAME") is not bound to any fonts', /home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/epaint-0.21.0/src/text/fonts.rs:624:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'Compute Task Pool (18)' panicked at 'A system has panicked so the executor cannot continue.: RecvError', /home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/bevy_ecs-0.10.0/src/schedule/executor/multi_threaded.rs:194:60
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/bevy_tasks-0.10.0/src/task_pool.rs:376:49
```

I added some code logging the existing `FontDefinitions` in the system showing the UI right before the crash, and it seems it just has some default fonts, `MY_FONT_NAME` is missing from it.

Were there any breaking changes to bevy_egui or egui that might reset `FontDefinitions` after I set them in my startup system and before they're being accessed in the UI system? Is there a better way to set custom fonts?

encounter:
`setup_new_windows_system` is added as both a `PreStartup` system and a `PreUpdate` system, and the `Query` will return the same `Added<Window>` for both. For window(s) created on startup, it will run again on the first frame and recreate the components, so any egui setup done with a startup system is lost.

One solution is to have the query omit any entities that already have an EguiContext.

Fixes #160

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -81,6 +81,7 @@ use bevy::{
     prelude::{
         Added, Commands, Component, CoreSet, Deref, DerefMut, Entity, IntoSystemAppConfigs,
         IntoSystemConfig, IntoSystemConfigs, Query, Resource, Shader, StartupSet, SystemSet, With,
+        Without,
     },
     render::{
         render_resource::SpecializedRenderPipelines, texture::Image, ExtractSchedule, RenderApp,
@@ -639,7 +640,10 @@ pub struct EguiManagedTexture {
 }
 
 /// Adds bevy_egui components to newly created windows.
-pub fn setup_new_windows_system(mut commands: Commands, new_windows: Query<Entity, Added<Window>>) {
+pub fn setup_new_windows_system(
+    mut commands: Commands,
+    new_windows: Query<Entity, (Added<Window>, Without<EguiContext>)>,
+) {
     for window in new_windows.iter() {
         commands.entity(window).insert((
             EguiContext::default(),


johanhelsing (approved):
Not the maintainer, but looks good to me :)

Fixes the regressions in my game with theme/font setup after upgrading from bevy 0.9 and beyv_egui 0.19.

mvlabat:
Great catch! Thank you for the fix

PR #166, opened 2023-03-21
move `bevy_core_pipeline` to dev-dependencies

jakobhellermann:
bevy_core_pipeline is only used in the render to texture example


mvlabat:
Great find, thank you!

> Would you be open to depending on bevy crates directly?

I was hesitant about depending on subcrates, as it makes patching more cumbersome for fork users. But compile times surely have higher priority, so if the gains are significant, I'm 100% open to this.

Issue #170, opened 2023-03-30
Accept KeyCode::NumpadEnter inputs


dimvoly:
Should hopefully be an easy one, when data entering into the various number fields (e.g. `DragValue`), it would be nice to be able to use the `KeyCode::NumpadEnter` to not take my hand away from the number pad.

For example, in the [egui demo](https://www.egui.rs/), in the circle Plot you can enter the r, x and y numbers with either the Enter key or the `NumpadEnter` key.

I believe all this needs is to have both of these redirect to Enter:

        KeyCode::Return => egui::Key::Enter,
        KeyCode::NumpadEnter => egui::Key::Enter,

in `systems.rs`.

dimvoly:
Think I'll do a PR for this, time I've learnt how to do a PR. PR request incoming soon.

PR #171, opened 2023-04-02
Accept NumpadEnter as Enter

dimvoly:
Just a single line of code added to be able to use NumpadEnter. See  [Issue #170](https://github.com/mvlabat/bevy_egui/issues/170).

To get this to compile I did have to change the level of `missing_docs` as the compiler complains of the `WorldQuery` derive macro being undocumented. See bevy [Issue #3492](https://github.com/bevyengine/bevy/issues/3492), the crate `bevy_ecs_macros` is not yet documented enough. Once that is fixed, then the level of documentation can go back up to `deny` (under a separate PR).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,4 +1,4 @@
-#![deny(missing_docs)]
+#![warn(missing_docs)]
 
 //! This crate provides a [Egui](https://github.com/emilk/egui) integration for the [Bevy](https://github.com/bevyengine/bevy) game engine.
 //!
--- a/src/systems.rs
+++ b/src/systems.rs
@@ -444,6 +444,7 @@ fn bevy_to_egui_key(key_code: KeyCode) -> Option<egui::Key> {
         KeyCode::Tab => egui::Key::Tab,
         KeyCode::Back => egui::Key::Backspace,
         KeyCode::Return => egui::Key::Enter,
+        KeyCode::NumpadEnter => egui::Key::Enter,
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![deny(missing_docs)]
+#![warn(missing_docs)]
```
mvlabat, line 1: Could you revert this change please? Other than that, looks good to me. Thanks!

mvlabat (changes requested):

src/lib.rs
mvlabat: On the second thought, I guess it'll make CI fail anyway. Let's keep it then :)

mvlabat (approved):

src/lib.rs
mvlabat: Oh I see. It seems like the new Bevy version breaks CI as its macro generates a struct without documentation.

dimvoly: We could `#![allow(missing_docs)]`? See [SO thread](https://users.rust-lang.org/t/solved-applying-attributes-to-macro-included-code/30484/7). Not too sure how to exclude a single derive macro from the missing docs, there's some [forum posts here](https://users.rust-lang.org/t/solved-applying-attributes-to-macro-included-code/30484/7), but it looks messy and might only work for proc macros.

Could allow and wait for bevy to write their docs.

Issue #113, opened 2022-07-22
copy to clipboard in wasm app


shnewto:
👋 Hello! I really love this crate, it's provided such an easy path to some of the things I'm working on in a small bevy app I've been working on lately! At the moment I've been working on adding some text to the system clipboard when a button is clicked and when I run the app with `cargo run` it does just that! But in the wasm version, nothing's making it to the clipboard when the button's clicked 🤔 

It seems like I'm able to reproduce in the bevy_egui_web_showcase app too.

https://user-images.githubusercontent.com/5607902/180486887-92f50f20-b8f4-4e07-8e2a-5927f98ddffd.mp4

Thanks for taking a look if you're able!

shnewto:
If it's helpful, I'm on macOS Monterey 12.4, and I've tested in Chrome, Safari, and Firefox and see the same behavior in each.

mvlabat:
Hi! Yeah, it's a known issue. It was a limitation on the `winit`'s end, but just about two weeks ago they merged a PR that's supposed to fix it: https://github.com/rust-windowing/winit/issues/1829
So thanks for creating the issue, I wouldn't check their progress otherwise :)

I imagine that the new `winit` version including those changes isn't released yet, and we'll also have to wait some time until it gets into Bevy itself.

frewsxcv:
This should get picked up in bevy if https://github.com/bevyengine/bevy/pull/5953 merges

frewsxcv:
...or https://github.com/bevyengine/bevy/pull/5347 or https://github.com/bevyengine/bevy/pull/6086

mvlabat:
I believe it's something that we can't do yet. While `winit` does expose the API to disable the `prevent_default` behaviour, there's no way to do that via Bevy yet. I guess it's something that should be exposed via `bevy_winit::window::WindowDescriptor`.

frewsxcv:
This is (I think) the relevant winit attribute that we need to set from bevy: https://github.com/rust-windowing/winit/blob/4b22ca8dafdbc5b17a32d489958b8f37eb1372de/src/platform_impl/web/window.rs#LL432C16-L432C27

johanhelsing:
Will be in Bevy 0.10: https://github.com/bevyengine/bevy/pull/7304

johanhelsing:
I also have a branch with the fix for bevy 0.9.1 if anyone wants to try it out with without depending on main https://github.com/johanhelsing/bevy/tree/optional_prevent_default

[...]

PR #178, opened 2023-04-18
web clipboard handling

Vrixyz:
Fixes https://github.com/mvlabat/bevy_egui/issues/113

- integrates https://github.com/rukai/cargo-run-wasm/blob/main/readme.md ; also used by winit to help with running examples easily ; could be done in a separate PR if need be.
- relies on events driven by web-sys to react to copy/paste events.
- Detects user agent to reliably detect mac. Probably better implemented in https://github.com/mvlabat/bevy_egui/pull/256
- On web, bevy 0.12 seems to have a bug, where qwerty keys are actually received. Seems to have been fixed by https://github.com/bevyengine/bevy/pull/10702 for bevy 0.13. I suggest we merge https://github.com/mvlabat/bevy_egui/pull/236 first then test this clipboard PR out.


Vrixyz:
Because the clipboard permissions are not implemented by all browsers, this is not an ideal...

More info: https://developer.mozilla.org/en-US/docs/Web/API/Permissions_API#browser_compatibility

On google chrome, as noted "supported on above link, I correctly got the permission request and my minimal playground for copy is working (shown in console) :tada:.

Buut having a permission request doesn't seem ideal, I feel like it's the wrong tool for the job ? I'm not sure about the exact process...

For me a past event should be handled by the browser, and forward characters to the canvas, winit, bevy, bevy_egui, egui....

I'm guessing `prevent_defaults()` is somehow in the way ? Any insight is welcome.

examples/clipboard.rs
```diff
@@ -0,0 +1,92 @@
+use bevy::{input::common_conditions::input_just_pressed, prelude::*};
+use bevy_egui::{egui, EguiContexts, EguiPlugin};
+use egui::{text_edit::CursorRange, TextEdit, Widget};
+use wasm_bindgen_futures::spawn_local;
+use web_sys::ClipboardPermissionDescriptor;
+
+fn main() {
+    App::new()
+        .add_plugins(DefaultPlugins)
```
paul-hansen, line 9: You can do this to turn off prevent_defaults now if that helps:
```suggestion
        .add_plugins(DefaultPlugins.set(WindowPlugin {
            primary_window: Some(Window {
                prevent_default_event_handling: false,
                ..default()
            }),
            ..default()
        }))
```

Vrixyz: I think I want to do that, I'm still digesting what exactly is the behaviour "default" that is referred to, how it helps with copypaste and why doesn't it work out of the box with that.

Maybe it's just the managed clipboard getting in the way, not sure..

paul-hansen: It prevents event handling on the html canvas Bevy uses from bubbling up to the rest of the page, which prevents hotkeys like F5 to refresh from working.

If your implementation of the clipboard requires JS to be handling the hotkeys, I think it would need to be set to false so you would get the ctrl+c/v events. I haven't looked too much at what you have in this PR, it might not be needed.

examples/clipboard.rs
```diff
@@ -56,37 +65,64 @@ fn clipboard_copy(mut text: ResMut<CustomText>) {
...
+
+    //        winit_window_instance.can
+    info!("setup_clipboard_paste OK");
+}
+
+fn read_clipboard_channel(mut clipboardChannel: NonSendMut<ClipboardChannel>) {
+    match &mut clipboardChannel.rx {
+        Some(rx) => {
+            if let Ok(clipboard_string) = rx.try_recv() {
+                info!("received: {}", clipboard_string);
```
Vrixyz, line 123: this doesn't get printed, I'm not sure why :(

examples/clipboard.rs
```diff
@@ -56,37 +65,64 @@ fn clipboard_copy(mut text: ResMut<CustomText>) {
...
+                info!("Not implemented.");
             }
-        };
+        }
+        info!("{:?}", event.clipboard_data())
     });
+
+    canvas
+        .add_event_listener_with_callback("paste", closure.as_ref().unchecked_ref())
+        .expect("Could not edd paste event listener.");
```
Vrixyz, line 111: I believe this callback doesn't work, I'm not 100% sure. ; I'm not sure how to debug in more details the exact stack trace when paste-ing.

Vrixyz:
9e85a902c51eb145929e6e52119d79c57380fb94 tries to implement `eframe`'s way of using web-sys API (following https://github.com/mvlabat/bevy_egui/issues/113#issuecomment-1537220739)

Vrixyz:
current state as of 69c447d2158cd4fbbbf3244272da0b68179d245a ; there is a weird bug causing the data pasted to be doubled, not sure why.

https://github.com/mvlabat/bevy_egui/assets/2290685/a3f29b72-6aee-416d-a840-2b6dfa41c8e2

examples/clipboard.rs
Vrixyz: https://github.com/mvlabat/bevy_egui/pull/178#discussion_r1195238342

examples/clipboard.rs
Vrixyz: Thanks, indeed after a few tests, prevent_default_handling has to be `false`  in order to get the events :tada: ; my other issue was that I was registering to "paste" events on the canvas, but I had to do it on the document.

Vrixyz: If you got any other information about this issue, I look forward to your feedbacks :)

src/web_clipboard.rs
Vrixyz, line 1: mahbe wasm32_clipboard would be a better name ?

Vrixyz:
I'm removing the "draft" mode, I think that's a fair attempt at adding that feature, I have a few weaknesses to point:
- should we remove completely the fully managed local "string" clipboard ? in wasm non web my implementation would fail 💀
  - I guess we could make some kind of decoration API with a bevy_egui_extra adding clipboard_web features...
- [eframe](https://github.com/emilk/egui/blob/master/crates/eframe/src/web/events.rs#L167) uses another egui API to support Paste, it's probably more correct because it's from official egui ; I didn't go this route because I fear other obstacles, I'd like to get a minimal clipboard support before imporving on top of that.

src/web_clipboard.rs
Vrixyz: not really because it uses web_sys...

src/systems.rs
```diff
@@ -250,20 +252,29 @@ pub fn process_input_system(
...
-                            focused_input.events.push(egui::Event::Copy);
-                        }
-                        egui::Key::X => {
-                            focused_input.events.push(egui::Event::Cut);
-                        }
-                        egui::Key::V => {
-                            if let Some(contents) = input_resources.egui_clipboard.get_contents() {
-                                focused_input.events.push(egui::Event::Text(contents))
+                {
+                    if command && pressed {
```
Vrixyz, line 256: ⚠️  on wasm mac, this is wrong, because it maps to ctrl key, but it should map to command key...

(to be clear, this is not a but introduced in this PR)

src/lib.rs
```diff
@@ -190,8 +202,8 @@ impl EguiClipboard {
 
     #[cfg(target_arch = "wasm32")]
     #[allow(clippy::unnecessary_wraps)]
-    fn get_contents_impl(&self) -> Option<String> {
-        Some(self.clipboard.clone())
+    fn get_contents_impl(&mut self) -> Option<String> {
+        self.clipboard.try_read_clipboard_event()
     }
```
mvlabat, line 207: That's a bit funny how reading from clipboard requires `&mut self`, but writing to it doesn't 😄

src/web_clipboard.rs
```diff
@@ -0,0 +1,106 @@
...
+
+fn setup_clipboard_paste(clipboard_channel: &mut WebClipboardPaste) {
+    let (tx, rx): (Sender<String>, Receiver<String>) = mpsc::channel();
+
+    use wasm_bindgen::closure::Closure;
+    use wasm_bindgen::prelude::*;
+
+    let closure = Closure::<dyn FnMut(_)>::new(move |event: web_sys::ClipboardEvent| {
+        // TODO: maybe we should check if current canvas is selected ? not sure it's possible,
+        // but reacting to event at the document level will lead to problems if multiple games are on the same page.
```
mvlabat, line 24: It should be pretty much possible, yeah. I'm ok if this isn't covered in this PR, but I guess the way to tackle it would look as follows:
1. Access a `winit` window via https://docs.rs/bevy/0.10.1/bevy/winit/struct.WinitWindows.html
2. Get the related canvas element: https://docs.rs/winit/latest/wasm32-unknown-unknown/winit/platform/web/trait.WindowExtWebSys.html#tymethod.canvas (it can be referenced as `web_sys::Element` too: https://docs.rs/web-sys/latest/web_sys/struct.HtmlCanvasElement.html#impl-AsRef%3CElement%3E-for-HtmlCanvasElement)
3. Compare it with the document's https://docs.rs/web-sys/latest/web_sys/struct.Document.html#method.active_element

src/web_clipboard.rs
```diff
@@ -0,0 +1,106 @@
...
+        rx: Some(Arc::new(Mutex::new(rx))),
+    };
+
+    info!("setup_clipboard_paste OK");
+}
+
+/// To get data from web paste events
+#[derive(Default)]
+pub struct WebClipboardPaste {
+    rx: Option<Arc<Mutex<Receiver<String>>>>,
```
mvlabat, line 58: I'd be happier if we used `NonSend` or crossbeam channels (supporting `Sync`) instead of a `Mutex`

src/web_clipboard.rs
```diff
@@ -0,0 +1,106 @@
...
+                None
+            }
+        }
+    }
+}
+
+/// Puts argument string to the web clipboard
+pub fn clipboard_copy(text: String) {
+    spawn_local(async move {
+        let window = web_sys::window().expect("window"); // { obj: val };
```
mvlabat, line 88: What does this comment mean?

src/systems.rs
mvlabat: Oh, good catch. This makes me wonder how we can detect running on MacOS in wasm.

Vrixyz: eframe uses events for all those, deferring the responsibility to the browser: https://github.com/emilk/egui/blob/307565efa55158cfa6b82d2e8fdc4c4914b954ed/crates/eframe/src/web/events.rs#L184

src/web_clipboard.rs
Vrixyz: not sure, a leftover, deleting.

Vrixyz, line 88:
```suggestion
        let window = web_sys::window().expect("window");
```

src/lib.rs
```diff
@@ -159,12 +163,20 @@ impl EguiClipboard {
...
+    /// Gets clipboard contents. Returns [`None`] if clipboard provider is unavailable or returns an error.
+    #[must_use]
+    #[cfg(target_arch = "wasm32")]
+    pub fn get_contents(&mut self) -> Option<String> {
+        self.get_contents_impl()
+    }
+
     #[cfg(not(target_arch = "wasm32"))]
-    fn set_contents_impl(&self, contents: &str) {
+    fn set_contents_impl(&mut self, contents: &str) {
```
Vrixyz, line 179: I'm not sure I intended that ? but it's probably weird to have different mutability depending on wasm on not

src/lib.rs
Vrixyz: yeah, I guess we could hide this by caching the web event, but that would require to:
- insert a system retrieving the web event (`try_recv`), store the result
- get that result here with a non mut access

I'm not sure about its benefits.

src/lib.rs
mvlabat: Yeah, we definitely should have a uniform interface. Let's keep everything `&mut self` then?

src/lib.rs
mvlabat: Yeah, I'm fine with keeping it `&mut self`, it was just an observation from my end :)

src/systems.rs
mvlabat: It feels like we should adopt the same approach. Let's keep the whole `match key` block enabled only for non-wasm targets, whereas for wasm we'll just read from channels (which in turn will get messages on web_sys events).

mvlabat: I think this is important to fix, as atm on MacOS, users need to do `Ctrl+C` to copy and `Cmd+V` to paste. I'd also be happy if we added the Select All (Ctrl/Cmd+A) support for WASM, to avoid the same inconsistency issue. That's extending the scope of this PR a bit, but hotkey consistency feels important here.

src/web_clipboard.rs
Vrixyz: I went for crossbeam channel, NonSend is a lifetime error fest and has to change egui res to NonSendMut

Cargo.toml
```diff
@@ -39,3 +42,24 @@ bevy = { version = "0.10", default-features = false, features = [
...
+    'EventTarget',
+    "Window",
+    "Navigator",
+] }
+js-sys = "0.3.63"
+wasm-bindgen = "0.2.84"
+wasm-bindgen-futures = "0.4.36"
+console_log = "1.0.0"
+log = "0.4"
+crossbeam-channel = "0.5.8"
```
Vrixyz, line 62: we might want to limit those dependencies only when we manage clipboard ? (or hotkeys events for select all ? (wip))

src/systems.rs
Vrixyz: - ctrl/cmd detection: this is how eframe does it, it checks the web_sys events for each keys... : https://github.com/emilk/egui/blob/307565efa55158cfa6b82d2e8fdc4c4914b954ed/crates/eframe/src/web/input.rs#LL211C25-L211C25

I feel like it means we would bypass winit/bevy key inputs in wasm :/ could this possibly fixed upstream 🤔 ?

src/web_clipboard.rs
Vrixyz: Created a follow up issue (a bit early but hopefully we'll manage to merge this PR)

https://github.com/mvlabat/bevy_egui/issues/185

src/web_clipboard.rs
```diff
@@ -0,0 +1,157 @@
...
+    let nav = window.navigator();
+    let platform = nav.platform();
+    match platform {
+        Ok(p) => info!(p),
+        Err(e) => ,
+    }
+    info!("{:?}", platform)
+}
+
+/// To get data from web paste events
```
Vrixyz, line 27:
```suggestion
/// To get data from web events
```

Cargo.toml
Vrixyz: actually now it's used to know if we're on mac or not.

src/lib.rs
```diff
@@ -138,6 +144,10 @@ impl Default for EguiSettings {
 #[derive(Component, Clone, Debug, Default, Deref, DerefMut)]
 pub struct EguiInput(pub egui::RawInput);
 
+/// A resource to check if we're on a mac, correctly detects on web too.
+#[derive(Resource)]
+pub struct IsMac(pub bool);
```
Vrixyz, line 149: maybe the name is a bit "common", EguiIsMac ? I don't know :)

src/web_clipboard.rs
```diff
@@ -0,0 +1,146 @@
...
+pub struct WebEventCut;
+/// Used asked to copy
+#[derive(Default)]
+pub struct WebEventCopy;
+
+fn setup_clipboard_copy(clipboard_channel: &mut WebChannel<WebEventCopy>) {
+    let (tx, rx): (Sender<WebEventCopy>, Receiver<WebEventCopy>) = crossbeam_channel::bounded(1);
+
+    let closure = Closure::<dyn FnMut(_)>::new(move |event: web_sys::ClipboardEvent| {
+        // TODO: maybe we should check if current canvas is selected ? not sure it's possible,
```
Vrixyz, line 52: we should probably remove these todos as there's an issue now

src/web_clipboard.rs
```diff
@@ -0,0 +1,146 @@
...
+        // but reacting to event at the document level will lead to problems if multiple games are on the same page.
+        tx.try_send(WebEventCopy);
+    });
+
+    // TODO: a lot of unwraps ; it's using documents because paste event set on a canvas do not trigger (also tested on firefox in vanilla javascript)
```
Vrixyz, line 57: any feeling about this ? should it fail silently ? with a warning/error log?

src/web_clipboard.rs
```diff
@@ -0,0 +1,146 @@
...
+        .unwrap()
+        .document()
+        .unwrap()
+        .add_event_listener_with_callback("cut", listener)
+        .expect("Could not add cut event listener.");
+    closure.forget();
+    *clipboard_channel = WebChannel::<WebEventCut> { rx: Some(rx) };
+}
+
+fn setup_clipboard_paste(clipboard_channel: &mut WebChannel<WebEventPaste>) {
```
Vrixyz, line 90: all those functions are mostly similar, that doesn't feel so good duplicating them, but there's only 3 and I feel any genericity would complicate their signature quite a bit

src/web_clipboard.rs
mvlabat: I'd change `unwrap` to `expect`, otherwise seems fine to me, as panicking on initialization failure doesn't seem to be silent :)

src/web_clipboard.rs
mvlabat: sounds good to me

src/lib.rs
```diff
@@ -532,7 +557,24 @@ impl Plugin for EguiPlugin {
...
         world.init_resource::<EguiUserTextures>();
         world.init_resource::<EguiMousePosition>();
+        if !world.contains_resource::<IsMac>() {
+            let mut is_mac = cfg!(target_os = "macos");
+            #[cfg(target_arch = "wasm32")]
+            {
+                let window = web_sys::window().expect("window");
+
+                let nav = window.navigator();
+                let platform = nav.platform();
```
mvlabat, line 567: Hm.. It looks like this property is deprecated: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/platform

src/web_clipboard.rs
mvlabat: yeah, fine by me 👍

Cargo.toml
```diff
@@ -39,3 +42,24 @@ bevy = { version = "0.10", default-features = false, features = [
     "bevy_pbr",
     "bevy_core_pipeline",
 ] }
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+winit = "0.28"
+web-sys = { version = "*", features = [
```
mvlabat, line 48: Let's specify a `web-sys` version too

.cargo/config.toml
```diff
@@ -0,0 +1,13 @@
+[alias]
+run-wasm = ["run", "--release", "--package", "run-wasm", "--"]
+
+# Credits to https://github.com/emilk/egui
+
+# clipboard api is still unstable, so web-sys requires the below flag to be passed for copy (ctrl + c) to work
+# https://rustwasm.github.io/docs/wasm-bindgen/web-sys/unstable-apis.html
+# check status at https://developer.mozilla.org/en-US/docs/Web/API/Clipboard#browser_compatibility
+#[build]
+#target = "wasm32-unknown-unknown"
```
mvlabat, line 10: We can drop these two lines I guess?

.cargo/config.toml
```diff
@@ -0,0 +1,13 @@
...
+
```
mvlabat, line 11:
```suggestion
# Using unstable APIs is required for writing to clipboard
```

src/web_clipboard.rs
```diff
@@ -0,0 +1,146 @@
...
+            .get_data("text/plain")
+        {
+            Ok(data) => {
+                tx.try_send(WebEventPaste(data));
+            }
+            _ => {
+                info!("Not implemented.");
+            }
+        }
+        info!("{:?}", event.clipboard_data())
```
Vrixyz, line 108: are these logs useful ? I suspect they would be noisy for end user.

src/lib.rs
Vrixyz: I went with `user_agent.contains("Macintosh;");`, I don't know if any user agent might include "Machine", I wanted to err on the side of precaution.

Vrixyz:
After more testing, I realize LWin and RWin is not properly forwarded from winit to bevy (to bevy_egui), I hope the winit keyboard input refactoring would fix it https://github.com/rust-windowing/winit/pull/2662 but I didn't look into it.

This results in cmd-A being unusable :( ; I can re-enable ctrl-A to avoid losing functionality, but as you said that creates a difference with cut/copy.

In my opinion, as a mac user, using `ctrl` feels very wrong, so I'd advocate to merge this PR with a fallback to `ctrl-A` while waiting or working on a fix upstream.

Vrixyz:
also, `prevent_default=false` (which I had to do to get copy/paste events) has other implications: cmd+left arrow triggers a "previous page", which eframe suppress via a `prevent_default = true` in a few other specific cases too.

I created an issue to winit to follow this prevent_default behaviour: https://github.com/rust-windowing/winit/issues/2831

Cargo.toml
```diff
@@ -39,3 +42,24 @@ bevy = { version = "0.10", default-features = false, features = [
...
+    'EventTarget',
+    "Window",
+    "Navigator",
+] }
+js-sys = "0.3.63"
+wasm-bindgen = "0.2.84"
+wasm-bindgen-futures = "0.4.36"
+console_log = "1.0.0"
+log = "0.4"
+crossbeam-channel = "0.5.8"
```
Vrixyz, line 62: maybe we should avoid specifying the last number (omit patch version) ?

mvlabat: That shouldn't matter much, as cargo won't prevent users of our crate doing semver compatible updates. I personally prefer omitting the patch part unless there's a hard requirement (e.g. no less than a specific patch version).

ConnorBP:
Any updates on this?

Vrixyz:
> Any updates on this?

Blockers listed on the description are still relevant ; I'm working on https://github.com/bevyengine/bevy/pull/8745.

If you want to help, I'd appreciate feedbacks on the problem I described here: https://github.com/Vrixyz/bevy/pull/3#discussion_r1396845228

extrawurst:
now that the winit update is merged into bevy can this go forward?

Vrixyz:
I believe so ; I'm still coordinating the follow up of winit update at  https://github.com/bevyengine/bevy/issues/11052 ; so I'm happy to let this adopted by someone else wants to ; otherwise I'll pick it up in a.. few month hopefully

Vrixyz:
My understanding has improved a bit.

My observations are base on an old version on bevy_egui, in order to be consistent with the state of this PR. this PR will need to be rebased anyway, and re-evaluated with winit changes. My intuition tells me that it should not too hard to update that PR.

# What problems this PR has

bevy_egui sends physical keys to egui, which is a problem on non qwerty keyboards, especially for "azerty select-all":

I thought the reason I wasn't receiving input from cmd-a was a bug from winit, but it was a wrong observation:
- as bevy_egui sends the physical keys, physical key  "A" will be sent when I hit the "Q" key, so to trigger select-all, I should do `cmd-Q`. but `cmd-Q` in mac is used to close an application. I remapped that to nothing because I don't want to hit that by mistake 😅 .
- Theoretically it should work, but winit doesn't send any event when typing `cmd-Q`. This might be a bug from upstream or the browser, but it's very niche as the correct use case should be to input `cmd-A`.

I think the correct approach is to fix bevy_egui to send the logical key rather than the physical key.

The winit update follow up (https://github.com/bevyengine/bevy/issues/11052) will help with detecting the correct "Logical key" with the task list "Add more info in KeyboardInput".

Given current state of bevy_egui, I'm thinking this does not create more inconsistencies in the API than currently the case on main, that being said, at this point I'm ok with ironing out the last bits before merging :+1: 

What would help greatly, for this PR and also future improvements, would be:
- a list of pertinent tests to do
- and better yet, automate them.

Vrixyz:
> Could you move forward with #236 I'd love to have this feature when 0.13 comes out. Thanks for all the winit work!

I would love to.

I'm starting to work on it: https://github.com/Vrixyz/bevy_egui/pull/3

As mentioned in https://github.com/mvlabat/bevy_egui/pull/178#issuecomment-1890458160 ; we still have the problem of bevy_egui using physical keys rather than logical keys on bevy **main**.

Those problems might not actually apply to bevy 0.12, and I now believe we could have merged that PR for bevy 0.12 if tested more thoroughly, but now I'll focus on making that working for bevy 0.13.

I did a talk on that subject for the curious, which might give more context: https://www.youtube.com/live/i3wF71XJ-24?si=zfFV8xoY9JuQKmOt&t=3090

src/web_clipboard.rs
Vrixyz: I removed most and/or changed for errors/warn

v-kat:
I might do a PR with a cleaned version of https://github.com/v-kat/bevy_egui/tree/wasm_mobile_keyboard it also seems to handle  some of the same copy paste functionality but I dunno if it covers everything (I'm mainly doing a hacky fix for the mobile web keyboard).

Vrixyz:
Great thanks for championing it to mergeability ! I'll let you close this PR at your will.

mvlabat:
@v-kat please do open a PR! I've just merged this one, so you can dissect the mobile web keyboard changes from the clipboard ones.

PR #179, opened 2023-04-27
Sampler descriptor moved into EguiSettings

GlummixX:
Hi, I've moved the sampler descriptor into the EguiSettings, since the hardcoded linear_descriptor is not desirable for pixel perfect usecases. I've also had to implement a compare_descriptors function to keep the PartialEq functionality. The sampler_descriptor can now be set to: 
- (Default) linear and clamping to edge
- (use_nearest_descriptor function) nearest and clamping to edge
- (use_bevy_descriptor function) derived from the bevy ImagePlugin setup
- completely custom

A simple demo. Left is linear, right is nearest.
image

### DIFF
--- a/src/egui_node.rs
+++ b/src/egui_node.rs
@@ -13,15 +13,14 @@ use bevy::{
     render::{
         render_graph::{Node, NodeRunError, RenderGraphContext},
         render_resource::{
-            AddressMode, BindGroupLayout, BindGroupLayoutDescriptor, BindGroupLayoutEntry,
-            BindingType, BlendComponent, BlendFactor, BlendOperation, BlendState, Buffer,
-            BufferAddress, BufferBindingType, BufferDescriptor, BufferUsages, ColorTargetState,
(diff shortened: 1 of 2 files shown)


mvlabat:
Hey. Thanks for the PR. I think we want to move this field to the plugin instance. Resources are meant for the settings that can be changed during runtime, but as far as I understand (please correct me if I'm wrong) this one is to be configured only once.

https://bevyengine.org/learn/migration-guides/0.8-0.9/#use-plugin-setup-for-resource-only-used-at-setup-time

GlummixX:
Hi, did some testing and it can be changed during runtime, the change will be visible after a cycle where is no egui rendered. Possible usecase is: nearest for pixel-perfect UI, while in game and linear for some different elements like showing images with lore. The bevy default cannot be changed after plugin setup. But the descriptor itself allows for changes.

mvlabat:
Ok, sounds good! And sorry for the late reply. Merged.

PR #181, opened 2023-05-22
Fix `@` on AZERTY on wasm

Vrixyz:
Follow up on https://github.com/mvlabat/bevy_egui/pull/149 ; I'm building on wasm and realized my fix is not sufficient, I can't input `@` on wasm :'(.

In https://github.com/mvlabat/bevy_egui/pull/149 we did a minimal change by checking only windows, but I think we're overly conservative, I'd like to discuss why this condition is needed, and make proper tests about it to validate its usefulness.

### DIFF
--- a/src/systems.rs
+++ b/src/systems.rs
@@ -210,15 +210,13 @@ pub fn process_input_system(
         }
     }
 
-    if !command || cfg!(target_os = "windows") && ctrl && alt {
-        for event in input_events.ev_received_character.iter() {
-            if !event.char.is_control() {
-                let mut context = context_params.contexts.get_mut(event.window).unwrap();
-                context
-                    .egui_input
-                    .events
-                    .push(egui::Event::Text(event.char.to_string()));
-            }
+    for event in input_events.ev_received_character.iter() {
+        if !event.char.is_control() {
+            let mut context = context_params.contexts.get_mut(event.window).unwrap();
+            context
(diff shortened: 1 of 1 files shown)


Vrixyz:
support for ctrl-A was mentionned, but I can attest this PR does not break it. (it's ctrl-Q on my keyboard though, but same behaviour as current live web showcase)

alice-i-cecile (commented):
What about ALT behavior? Shouldn't those keys also be skipped, like in the original implementation?

Vrixyz:
I think we actually don't want to prevent ALT, with this PR it enables me to input weird characters with altcodes (alt + keys): like the `alt` + (`1`;`3`;`0`) inputs `é` (https://www.alt-codes.net/)

alice-i-cecile (approved):

Vrixyz:
> I'm afraid this breaks the `Cmd+A` functionality on MacOS: selecting the text works, but it also inserts the "A" character afterwards.

Now I believe that's because we need to accurately differentiate ctrl from command, from the clipboard PR: https://github.com/mvlabat/bevy_egui/pull/178/files#diff-420f4e08a25299e0e1fb09f4757e1d5d027a2278f37030c09ab9c06786cfa52eR85-R99

Vrixyz:
I'm closing this and will work on previous comment

PR #182, opened 2023-05-28
Bump version, update egui/bevy dependency version

dymk:
- Updates Bevy dependency to be the latest @ 0.10.1
- Updates Egui dependency to be the latest @ 0.22.0

And bump the crate version to 0.20.4, as the change as not backwards compatible for projects pinned at different bevy/egui versions. 

Perhaps there's a better way to make bevy_egui agnostic to specific bevy/egui versions? They're pre-1.x.y, so semver doesn't make compatibility guarentees, but I'm hoping the crates don't have to be pinned at the patch version.


mvlabat:
Hey! Thank you for the PR. I'll close this in favour of #184, as it has conflicts in `Cargo.toml` after updating to Bevy 0.11 now.
My apologies for not checking it sooner.

PR #188, opened 2023-06-18
Update to bevy 0.11

Vrixyz:


src/lib.rs
```diff
@@ -534,76 +534,76 @@ impl Plugin for EguiPlugin {
...
-                .in_base_set(CoreSet::PostUpdate),
+                .after(EguiSet::ProcessOutput),
         );
-        app.add_system(free_egui_textures_system.in_base_set(CoreSet::Last));
+        app.add_systems(Last, free_egui_textures_system)
+                .add_systems(Render,
+                    render_systems::prepare_egui_transforms_system.in_set(RenderSet::Prepare),
+                )
+                .add_systems(Render, render_systems::queue_bind_groups_system.in_set(RenderSet::Queue))
+                .add_systems(Render, render_systems::queue_pipelines_system.in_set(RenderSet::Queue));
```
Vrixyz, line 579: those "Render" surprised me a bit.

src/lib.rs
```diff
@@ -534,76 +534,76 @@ impl Plugin for EguiPlugin {
...
 
         let mut shaders = app.world.resource_mut::<Assets<Shader>>();
         shaders.set_untracked(
             EGUI_SHADER_HANDLE,
             Shader::from_wgsl(include_str!("egui.wgsl")),
         );
+    }
 
+    fn finish(&self, app: &mut App) {
```
Vrixyz, line 588: not sure if necessary, I did that because I had a crash due to wrongly using `Update` instead of `Render` ; and noticed the post processing example did initialize its resources in `finish` rather than `build` ; probably a better practice to allow other plugins to prewrite the resource ? not sure though, please challenge :)

src/systems.rs
```diff
@@ -143,12 +143,11 @@ pub fn process_input_system(
         // that has been left.
         if cursor_left_window != Some(cursor_moved.window) {
             let scale_factor = egui_settings.scale_factor as f32;
-            let mut mouse_position: (f32, f32) = (cursor_moved.position / scale_factor).into();
+            let mouse_position: (f32, f32) = (cursor_moved.position / scale_factor).into();
             let mut context = context_params
                 .contexts
                 .get_mut(cursor_moved.window)
                 .unwrap();
-            mouse_position.1 = context.window_size.height() / scale_factor - mouse_position.1;
```
Vrixyz, line 151: noticed the axis was now inverted so I had to remove that.

src/lib.rs
mvlabat: Sounds good to me

Cargo.toml
```diff
@@ -42,3 +42,6 @@ bevy = { version = "0.11", default-features = false, features = [
     "bevy_pbr",
     "bevy_core_pipeline",
 ] }
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+bevy = { version = "0.11", features = ["webgl2"] }
```
Vrixyz, line 47: I guess the alternative is to target webgpu, but as it's mostly experimental, it should be fine ? Or we expose this control through a feature flag so user can choose 🤔

Vrixyz: solved by updating ci to use unstable web_sys api (hopefully)

seabassjh:
The simple demo in WASM is broken (black screen) but I found a fix for it!
https://github.com/Vrixyz/bevy_egui/pull/1/files

Everything else works great :)

mvlabat:
Huge thanks @Vrixyz! I'm merging this and will hopefully make a release soon.
@seabassjh thanks, I'll apply your fix too.

Issue #145, opened 2023-01-17
IMEs not working 


liquidhelium:
I tried the demo and found I can't input non-Latin characters since the IME just won't show up.  
In egui native & web demos, IME works fine.  
platform: Windows, Edge browser (chromium)

doup:
#149 doesn't fix this. But, I think after `0.10` is released this weekend it'll be possible to fix it with: https://github.com/bevyengine/bevy/pull/7325 (see `text_input` example in `bevy/main`).

`bevy_egui` or the user will need to set:

```
window.ime_enabled = true;
window.ime_position = ...;
```

It probably makes sense to do this in `bevy_egui` so that `ime_position` is set to the active input field.

yangzhixiao:
Very concerned about this issue.

PR #204, opened 2023-08-24
Basic IME mapping

EReeves:
Fixes https://github.com/mvlabat/bevy_egui/issues/145

This is enough for my needs, but FYI:
- `Ime::Disabled` just maps to `Event::CompositionEnd` with an empty string for text. You could capture `Ime::Preedit` as state to handle it that way. I don't know where best to add the state.

### DIFF
--- a/src/systems.rs
+++ b/src/systems.rs
@@ -22,7 +22,7 @@ use bevy::{
     log::{self, error},
     prelude::{Entity, EventReader, NonSend, Query, Resource, Time},
     time::Real,
-    window::{CursorMoved, RequestRedraw},
+    window::{CursorMoved, Ime, RequestRedraw},
     winit::{EventLoopProxy, WakeUp},
 };
 use std::{marker::PhantomData, time::Duration};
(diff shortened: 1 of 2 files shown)


src/systems.rs
```diff
@@ -231,6 +233,39 @@ pub fn process_input_system(
...
+            .egui_input
+            .events
+            .push(event);
+    }
+
+    for ev in input_events.ev_ime_input.iter() {
+        match ev {
+            Ime::Preedit {
+                window,
+                value: _,
```
mvlabat, line 250: shouldn't we also push a `CompositionUpdate` event with this value right after pushing `CompositionStart`?

(looking at the reference implementation at https://github.com/emilk/egui/blob/2bc2fb9c393bbbfc4832a35af1c100bc4ea48719/crates/egui-winit/src/lib.rs#L301)

mvlabat:
Hey! Thank you for the PR and apologies for taking a while to review it. Looks good to me in general, left 1 comment about the `Ime::Preedit` event though

src/systems.rs
EReeves: Sounds good to me. I read through the reference and all code that uses the event. The event is used by TextEdit to preview text before it's committed by the IME. It should also be added for the sake of completion.

liquidhelium (approved):

liquidhelium (approved):

src/systems.rs
mvlabat: Sorry for forgetting to comment on this again, I feel really bad for neglecting this PR for several releases. The last question I want to ask is if there is a reason why we don't align our implementation to the reference one completely.

I haven't had a chance to test it myself and to understand if there's really any difference in behaviour, but I guess I'd feel safer if we just adopted exactly the same logic.

v-kat:
Will this get merged soon?

v-kat:
Is this related to WASM mobile keyboard display at all? I've been trying to debug it and had assumed it was an winit issue or maybe something with IME (everything works fine with just egui).

PR #206, opened 2023-09-09
Update bevy support table in readme

lukors:


mvlabat:
Hey, I think your fork is a bit outdated, our readme already has that entry in the table.

lukors:
Ahh, sorry, I had GitHub set to an old version. xD

PR #208, opened 2023-09-26
Move components and resources to module

TheButlah:
It was a bit hard to keep track of all the datatypes, so I moved all the non-render components and resources to the `components` module


TheButlah:
Closing after discussion on discord - opting to prefer components co-located near the systems that use them

PR #214, opened 2023-10-02
Add worldspace egui

TheButlah:
Rebased on #210 and #211.

Adds worldspace UI to `bevy_egui` via the `EguiRenderToTexture` component. Handling input is not implemented in this PR.

This PR is marked as a draft, because I thought that I could have multiple StandardMaterials on the same mesh, but that is not the case right now in bevy. This means that a user cannot easily have a base texture and then apply the egui overlaid transparently on top.

I think the right solution is likely to include an `EguiMaterial` as part of this plugin for users' convenience, and keep the `EguiRenderToTexture` as an option for people to use if they want to create their own custom materials. I don't have any experience with bevy materials yet, so I will revisit that later.

I'm also not particularly happy with my refactor - I feel like the systems could be unified more - there is still some branching logic to handle windows vs textures.

image

### DIFF
--- a/src/render_systems.rs
+++ b/src/render_systems.rs
@@ -1,6 +1,6 @@
 use crate::{
     egui_node::{EguiNode, EguiPipeline, EguiPipelineKey},
-    EguiManagedTextures, EguiSettings, EguiUserTextures, WindowSize,
+    EguiManagedTextures, EguiRenderToTexture, EguiSettings, EguiUserTextures, WindowSize,
 };
 use bevy::{
     ecs::system::SystemParam,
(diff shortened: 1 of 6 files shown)


TheButlah:
Hi, @mvlabat what do you think I should take as the next steps for getting this unblocked?

src/render_systems.rs
```diff
@@ -21,9 +23,16 @@ use bevy::{
     utils::HashMap,
 };
 
-/// Extracted Egui settings.
-#[derive(Resource, Deref, DerefMut, Default)]
-pub struct ExtractedEguiSettings(pub EguiSettings);
+/// The extracted version of [`EguiManagedTextures`].
+#[derive(Debug, Resource)]
+pub struct ExtractedEguiManagedTextures(pub HashMap<(Entity, u64), Handle<Image>>);
```
mvlabat, line 28: Let's turn this `(Entity, u64)` tuple into a struct, named fields will help to understand what `Entity` and `u64` are related to.

mvlabat:
@TheButlah apologies for the delayed response. I'm yet to take a closer look to better answer the question. And I'll need to re-read our Discord DM to remember all the context :)
Anyway, as for the first step, let's rebase this against the latest Egui version

TheButlah:
> Anyway, as for the first step, let's rebase this against the latest bevy_egui version

Just completed this :) However, the cube now looks like this:
Screenshot 2024-01-09 at 1 40 53 AM

It turns out, this is identical to what I get when I `cargo run --example worldspace` in my [original commit](https://github.com/TheButlah/bevy_egui/commit/6e583f9ac2c792b552a941a06f097f76c9328dc0). IDK then if the above screenshot came from some uncomitted tweak I did or what, but it never actually did the transparency correctly, I guess.

When I was doing my experiments with wgpu directly, this would happen when the `BlendState` was not quire configured correctly, causing it to repeatedly add until the values were opaque instead of partially transparent. I'm not sure exactly if that is what is happening here.

I'll need to revise the PR to fix it, but feel free to fix it yourself if you know how!

v-kat:
Are you going to convert this to not a draft? It looks very useful....

mvlabat:
Merged as part of #304. Thank you so much for your initial work on this!

PR #220, opened 2023-10-22
Add `add_srgb_suffix` to fix #194 for WASM

### LINKED ISSUE #194: Incompatible color attachments under WebGPU/WASM
The color attachment state expects SRGB in WebGPU.  Adding it to the texture_format in `render_systems.rs` fixes this issue. 

```                
texture_format: window.swap_chain_texture_format?.add_srgb_suffix(),
```

It is unclear, however, if applying this fix unconditionally to all platforms would be a breaking change or not.  What is clear, is that this fix at least works with WebGPU.

image

I know @mvlabat and I discussed this in the bevy discord already, I just wanted to make sure this issue does not fall through the cracks

frewsxcv:
This fixes https://github.com/mvlabat/bevy_egui/issues/194 for me

### DIFF
--- a/src/render_systems.rs
+++ b/src/render_systems.rs
@@ -224,7 +224,7 @@ pub fn queue_pipelines_system(
         .iter()
         .filter_map(|(window_id, window)| {
             let key = EguiPipelineKey {
-                texture_format: window.swap_chain_texture_format?,
+                texture_format: window.swap_chain_texture_format?.add_srgb_suffix(),
             };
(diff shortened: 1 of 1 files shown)


mvlabat:
I didn't see any difference in colours on both MacOS and Windows after applying the change, so I guess it's safe to merge. Thanks for the PR!

Hellzbellz123:
not only does this fix that issue, it fixes a crash on nvidia and wayland for me :)

PR #221, opened 2023-10-23
upgrade to bevy 0.12

raffaeleragni:
to update cargo when bevy 0.12 is out

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -93,7 +93,7 @@ use bevy::{
     utils::HashMap,
     window::{PrimaryWindow, Window},
 };
-use std::borrow::Cow;
+use std::{borrow::Cow, mem::discriminant};
 #[cfg(all(feature = "manage_clipboard", not(target_arch = "wasm32")))]
 use std::cell::{RefCell, RefMut};
 #[cfg(all(feature = "manage_clipboard", not(target_arch = "wasm32")))]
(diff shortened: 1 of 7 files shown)


frewsxcv:
Bevy 0.12 was just released!

raffaeleragni:
yea, cargo updated

frewsxcv (approved):

src/lib.rs
```diff
@@ -144,7 +144,7 @@ fn compare_descriptors(a: &ImageSampler, b: &ImageSampler) -> bool {
     match (a, b) {
         (ImageSampler::Default, ImageSampler::Default) => true,
         (ImageSampler::Descriptor(descriptor_a), ImageSampler::Descriptor(descriptor_b)) => {
-            descriptor_a == descriptor_b
+            descriptor_a.label == descriptor_b.label
```
paul-hansen, line 147: Idk if this might cause issues. Labels are optional so if multiple descriptors don't provide a label (or provide the same label) this could cause false equality checks. 
Not sure what the solution should be, we could push for bevy to implement partial eq like wgpu has or we could try to manually implement it. 
Also depending on what relies on this function, maybe it isn't important enough to worry about or is always used for descriptors we control and know are unique labels?

raffaeleragni: compared all of them

src/lib.rs
```diff
@@ -143,8 +143,18 @@ impl PartialEq for EguiSettings {
...
+                && discriminant(&a.address_mode_u) == discriminant(&b.address_mode_u)
+                && discriminant(&a.address_mode_v) == discriminant(&b.address_mode_v)
+                && discriminant(&a.address_mode_w) == discriminant(&b.address_mode_w)
+                && discriminant(&a.mag_filter) == discriminant(&b.mag_filter)
+                && discriminant(&a.min_filter) == discriminant(&b.min_filter)
+                && discriminant(&a.mipmap_filter) == discriminant(&b.mipmap_filter)
+                && a.lod_min_clamp == b.lod_min_clamp
+                && a.lod_max_clamp == b.lod_max_clamp
+                && a.anisotropy_clamp == b.anisotropy_clamp
+                && discriminant(&a.border_color) == discriminant(&b.border_color)
```
ManevilleF, line 157: Shouldn't this be moved in a `Eq`/`PartialEq` impl ?

src/systems.rs
```diff
@@ -358,7 +358,7 @@ pub fn process_input_system(
     }
 
     for mut context in context_params.contexts.iter_mut() {
-        context.egui_input.predicted_dt = time.raw_delta_seconds();
+        context.egui_input.predicted_dt = time.delta_seconds();
```
ManevilleF, line 361: Would that change have unwanted effects? It was `raw_delta_seconds` , meaning it should use `Time<Real>`, and yet I think this system uses virtual time

src/lib.rs
raffaeleragni: those types are not owned by the crate

src/systems.rs
raffaeleragni: no idea

mvlabat:
@raffaeleragni thanks a lot for the PR! There are a few CI warnings, but I'll take it from here. (I'll push directly to your branch, I hope you won't mind.)

raffaeleragni:
no prob

src/systems.rs
czaloj: Function signature should be updated to use Time<Real>

raffaeleragni: ok

src/lib.rs
mvlabat: This seems to work:
```
        (ImageSampler::Descriptor(descriptor_a), ImageSampler::Descriptor(descriptor_b)) => {
            descriptor_a.as_wgpu() == descriptor_b.as_wgpu()
        }
```

(I've pushed the updated version)

PR #226, opened 2023-11-09
Handle giving time input to egui correctly

TheRawMeatball:
The predicted_dt should be based on the target framerate and not depend on frame-to-frame fluctuations. Instead, we can pass total elapsed time instead to let egui handle things correctly.


mvlabat:
Hi, apologies for the delay, and thanks for the fix!

PR #237, opened 2023-12-23
Add render feature to allow crates use bevy_egui that do not use bevy_render feature

BeastLe9enD:
I added a `render` feature that is enabled by default to allow applications that want to use `bevy_egui` (but are not using `bevy_render` because they have a own rendering plugin), do not need to fully reimplement `bevy_egui` on their own and just disable the `bevy_render` dependent stuff :)


mvlabat:
Thanks for the PR! Merged as part of #243

PR #238, opened 2023-12-29
Integrate AccessKit

ndarilek:
This integrates Bevy's AccessKit setup with egui's AccessKit integration so egui UIs are accessible.

I haven't been able to test this on Linux, where it may not work since accesskit integration is feature-gated there. In theory it should just pull in dependencies and more or less be a no-op since an access technology will never request accessibility, but I can't say for sure and would appreciate help testing.

Thanks!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1043,9 +1043,12 @@ impl Plugin for EguiPlugin {
 
         app.add_systems(
             PostUpdate,
-            update_accessibility
-                .after(EguiPostUpdateSet::ProcessOutput)
-                .before(AccessibilitySystem::Update),
+            update_accessibility.in_set(EguiPostUpdateSet::PostProcessOutput),
+        );
+
+        app.configure_sets(
+            PostUpdate,
+            EguiPostUpdateSet::PostProcessOutput.before(AccessibilitySystem::Update),
         );
     }
 
@@ -1117,15 +1120,13 @@ pub fn setup_new_windows_system(
     for window in new_windows.iter() {
         let context = EguiContext::default();
         if let Some(adapters) = &adapters {
-            println!("Got adapters");
-            if  adapters.get(&window).is_some() {
-                println!("Got a window");
+            if adapters.get(&window).is_some() {
                 context.ctx.enable_accesskit();
                 **manage_accessibility_updates = false;
             }
         }
         // See the list of required components to check the full list of components we add.
-        commands.entity(window).insert(EguiContext::default());
+        commands.entity(window).insert(context);
     }
 }
 
@@ -1494,9 +1495,6 @@ pub fn end_pass_system(
     for (mut ctx, egui_settings, mut full_output) in contexts.iter_mut() {
         if !egui_settings.run_manually {
             **full_output = Some(ctx.get_mut().end_pass());
-            if full_output.0.as_ref().unwrap().platform_output.accesskit_update.is_some() {
-                println!("Got one!");
-            }
         }
     }
 }
@@ -1508,11 +1506,8 @@ fn update_accessibility(
     mut adapters: NonSendMut<AccessKitAdapters>,
 ) {
     if requested.get() {
-        println!("Accessibility requested");
         for (entity, output) in &outputs {
-            println!("Got an output");
             if let Some(adapter) = adapters.get_mut(&entity) {
-                println!("Got  an adapter");
                 if let Some(update) = &output.platform_output.accesskit_update {
                     **manage_accessibility_updates = false;
                     adapter.update_if_active(|| update.clone());
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -121,6 +121,7 @@ pub mod text_agent;
 #[cfg(all(feature = "manage_clipboard", target_arch = "wasm32",))]
 pub mod web_clipboard;
 
+use bevy_a11y::{AccessibilityRequested, AccessibilitySystem, ManageAccessibilityUpdates};
 pub use egui;
 
 use crate::input::*;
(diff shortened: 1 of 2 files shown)


mvlabat:
@ndarilek thank you very much for this PR!

Could you give me some hints about how I can test it on MacOS or Windows? UI accessibility is a really new topic for me, but I'll be glad to become more educated about it.

src/lib.rs
```diff
@@ -1078,8 +1086,19 @@ pub struct EguiManagedTexture {
 pub fn setup_new_windows_system(
     mut commands: Commands,
     new_windows: Query<Entity, (Added<Window>, Without<EguiContext>)>,
+    adapters: Option<NonSend<AccessKitAdapters>>,
+    mut manage_accessibility_updates: ResMut<ManageAccessibilityUpdates>,
 ) {
     for window in new_windows.iter() {
+        let context = EguiContext::default();
+        if let Some(adapters) = &adapters {
+            println!("Got adapters");
```
vladbat00, line 1095: Not sure if it's ready for review or not, but we should get rid of `println!` statements (or we can replace them with `log::debug!`)

vladbat00: ah, nvm, didn't notice the "WIP" tag in the commit :)
but feel free to leave the ones you can find useful for debugging as `log::debug`

ndarilek: Oh heh, totally forgot I'd opened a PR for this. No worries.

Yes, thanks for flagging the `println`s, I needed something to output directly to my screen reader without the rest of the debug/info-level logging coming along with it. :) Will definitely remove those from the final PR.

Does the rest look OK from the egui side, though? In particular, I didn't know if I needed to copy `accesskit_update` from egui differently. It definitely won't work under 0.15 due to [this](https://github.com/bevyengine/bevy/pull/18346) but I'm still not entirely clear on how/where properties are picked up from egui. Hopefully I got it right, I'd like to merge it early in the 0.16 release cycle if possible.

Thanks.

vladbat00: Cool, ty! :)

> Does the rest look OK from the egui side, though?

Yeah, there's something in the code I'm not so sure about.

Here, you've added creation of new `EguiContext` for which you enable accesskit, but then the context isn't passed anywhere, so it just gets destroyed when it leaves the scope.

You probably want to insert the `context` variable to the window entity:
`commands.entity(window).insert(context)`.

> I didn't know if I needed to copy `accesskit_update` from egui differently.

Looks good to me so far, even though I don't particularly know about Bevy's `ManageAccessibilityUpdates`, so I can't confirm that with confidence. 🙃 

Another thing I'd probably like this PR to add is a new example (if that makes sense). I assume activating accessibility features requires some setup in the user code as well? If so, that's something for which an example would be useful.

I'd also appreciate if you could provide some instructions on how to test this on Windows or MacOS. Maybe they can go as comments in the example code. I'd be happy to play around with it and test it, but I barely know anything about accessibility features unfortunately.

src/lib.rs
```diff
@@ -1014,6 +1015,13 @@ impl Plugin for EguiPlugin {
             "egui.wgsl",
             bevy_render::render_resource::Shader::from_wgsl
         );
+
+        app.add_systems(
+            PostUpdate,
+            update_accessibility
```
vladbat00, line 1021: Let's add this system to the `EguiPostUpdateSet::PostProcessOutput` set, and configure the whole set to run `before(AccessibilitySystem::Update)`.

src/lib.rs
```diff
@@ -1450,6 +1469,32 @@ pub fn end_pass_system(
...
+                println!("Got one!");
+            }
+        }
+    }
+}
+
+fn update_accessibility(
+    requested: Res<AccessibilityRequested>,
+    mut manage_accessibility_updates: ResMut<ManageAccessibilityUpdates>,
+    outputs: Query<(Entity, &EguiOutput)>,
```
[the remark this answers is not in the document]
vladbat00: Fixed in 83e3fb4

ndarilek:
I believe I've addressed all your PR feedback and, thanks to your catch of me failing to add the context to the window, this now works under Linux! (It wasn't before and I thought it might be a Bevy issue.)

There's not really anything someone has to know/do to use accessibility support, so I don't think it needs any specific examples. Egui widgets just work automatically. I'd like to add additional accessibility helpers, but those would be in egui itself so documented there.

Testing is a bit weird right now. It works under Linux/Orca and maybe macOS/VoiceOver. VO is a bit complicated to use and I don't regularly, but I think that pressing cmd-f5 should bring it up. There are lots of review commands, but you should be able to at least tab through the interface, arrow around controls, etc.

Unfortunately Windows won't work at all until Bevy 0.16 is released due to the above-linked PR--Windows is a bit finicky about how its accessibility tree is created and that was broken in 0.15. That's why I'm a bit uncertain about macOS--I *think* the 0.15 breakage is specific to Bevy but I don't have a mac at hand to test.

I know there are accessible testing tools that let you introspect the accessibility tree so you can see the structure without having to run a screen reader. I don't remember what those are right now but I'll look tomorrow.

Thanks for all the feedback.

vladbat00:
I had to push some changes due to merge conflicts and also fixed some little things around system configs and made them public.
Just tested it on MacOS, works great, thank you! (Sometimes it says "*app* is not responding", but it just takes a couple of seconds to start reacting to the events.)

vladbat00:
I've merged the PR, but I'll have to disable the functionality for some time unfortunately, as there's an `accesskit` version incompatibility between Bevy 0.16 and the current version of Egui. We'll have to wait for the next Egui release to be able to re-enable it.

ndarilek:
Understood, getting the various bevy/egui/accesskit/winit versions to line up on their respective release cycles has definitely been a pain point. Thanks for at least merging it--good to not be maintaining this PR on my own. :)

PR #240, opened 2024-01-03
Fix requested repaints not causing Bevy to redraw.

### LINKED ISSUE #239: Using WinitSettings::desktop_app() does not update egui UI
I'm using `WinitSettings::desktop_app()` in my bevy app (so the window is only updated when a winit event is received or a `RequestRedraw` event is sent manually), but it seems that `RequestRedraw` events aren't sent by `bevy_egui` when the UI needs to be updated. This results in behaviour like this:

https://github.com/mvlabat/bevy_egui/assets/98399119/3df9eef8-5742-47b9-9618-ea15de96779a

(although I click the collapsing header, it only collapses when I wiggle the mouse because that's when the next Winit event is received).

I tried to add this bevy system to the `PostUpdate` schedule to fix the problem:
```rust
fn redraw_window(mut contexts: EguiContexts, mut ev_request_redraw: EventWriter<RequestRedraw>) {
    let ctx = contexts.ctx_mut();
    // redraw the window if egui wants the window redrawn
    if ctx.input(|i| i.wants_repaint()) {
        ev_request_redraw.send(RequestRedraw);
    }
}
```
And it did partially fix it, as you can see in this screen recording:

https://github.com/mvlabat/bevy_egui/assets/98399119/8349dddc-e6ea-4680-8c9b-b2bd559a114d

But the window is still not being redrawn enough times. 

Do you have any suggestions on getting this to work manually, or is there a chance this feature could be added to `bevy_egui`?
Thanks

andriyDev:
The previous implementation seems to have lost track of what the redraw was supposed to be tracking. This meant button clicks would **not** trigger a redraw on `WinitSettings::desktop_app()` (only a mouse move after would cause a redraw).

I've tested this on my own project and things work much more correct!

Fixes #239.

### DIFF
--- a/src/systems.rs
+++ b/src/systems.rs
@@ -428,7 +428,7 @@ pub fn process_output_system(
             shapes,
             textures_delta,
             pixels_per_point,
-            viewport_output,
+            viewport_output: _,
         } = full_output;
         let paint_jobs = ctx.tessellate(shapes, pixels_per_point);
 
--- a/src/systems.rs
+++ b/src/systems.rs
@@ -458,7 +458,7 @@ pub fn process_output_system(
         #[cfg(not(windows))]
         set_icon();
 
-        if viewport_output.is_empty() {
+        if ctx.has_requested_repaint() {
             event.send(RequestRedraw)
         }
 


aevyrie:
It looks like the correct solution is to use this field from egui's `FullOutput`: https://docs.rs/egui/latest/egui/viewport/struct.ViewportOutput.html#structfield.repaint_delay
and find if any viewport wants a repaint. Even better, use bevy's winit proxy to schedule the exact wakeup time based on the minimum `repaint` delay.

It appears this was mangled during the last release.

aevyrie (approved):
Although I noted another "better" solution, this change is an improvement that fixes a regression, and matches the original implementation.

andriyDev:
> It looks like the correct solution is to use this field from egui's `FullOutput`: https://docs.rs/egui/latest/egui/viewport/struct.ViewportOutput.html#structfield.repaint_delay and find if any viewport wants a repaint. Even better, use bevy's winit proxy to schedule the exact wakeup time based on the minimum `repaint` delay.

As far as I can tell, viewports don't actually know if they need a redraw or not. As in "request redraw" only applies to the whole context (so all viewports). That said I agree that the best solution also takes into account the redraw_delay as you noted.

My point being: this PR will not need to be reverted or rewritten, just more code will need to be added to fully handle all the cases.

mvlabat:
Thank you for the fix! I'm good to merge it as is, at least that will fix the immediate regression. If anyone wants to submit a PR with a proper fix, I'll gladly accept it as well (I don't have the time to look into that myself, unfortunately).

PR #241, opened 2024-01-05
disable manage clipboard for wasm32 AND android

### LINKED ISSUE #198: arboard fails to build for android target (and probably iOS)
Obviously, if *bevy_egui* is a direct dependency, you can work around this with:
```toml
[dependencies]
bevy_egui = { version = "0.21.0", default-features = false, features = ["open_url", "default_fonts"] }
```
However, if you are targeting multiple platforms, it gets more complicated:
```toml
[target.'cfg(target_os = "android")'.dependencies]
bevy_egui = { version = "0.21.0", default-features = false, features = ["open_url", "default_fonts"] }

[target.'cfg(not(target_os = "android"))'.dependencies]
bevy_egui = { version = "0.21.0"  }
```
And, if it happens to be a dependency of third party plugins you use (such as bevy_console) and isn't feature-controlled there, I think you are just out of luck unless you're prepared to fork all of those projects.

Given that there's already an explicit exclusion for wasm, it seems reasonable to extend it to something like:
```toml
[target.'cfg(not(any(target_arch = "wasm32", target_os = "android", target_os = "ios")))'.dependencies]
```

I would guess the same is true of iOS so I stuck that condition in as well but I don't have a Mac to test with and I haven't forked bevy_egui so that last block is entirely untested.

Though, really, it'd be nice to have clipboard support on Android (and iOS) at some point. It's not something I have the time or expertise to take on at the moment but maybe in the future.

Hellzbellz123:
Closes #198 (and the duplicate I was about to make :0).

arboard dependency does not support android and fails too build for said platform,

this pr modifies the existing target_arch checks too include target_os = "android" check

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -82,6 +82,73 @@ jobs:
             cache-wasm32-cargo
       - run: cargo clippy --no-default-features --target=wasm32-unknown-unknown --all-targets --features=${{ matrix.features }} -- -D warnings
 
+  clippy_android:
+    name: Clippy check (android)
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        features:
+          [
+            "immutable_ctx",
+            "manage_clipboard",
+            "open_url",
+            "manage_clipboard,open_url",
+          ]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@master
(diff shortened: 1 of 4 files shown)


mvlabat:
Thank you for the fix! I made some changes to the CI workflows to cover the mobile platforms as well, that should help to keep the crate compilable for them as well.

PR #248, opened 2024-01-09
Update to egui 0.25.0

MickHarrigan:
The only major thing of note here is in `src/systems.rs` where a new `Key` field is introduced that takes both the key that is pressed and a physical key that allows non-qwerty users to still use the same layout.

I don't know if this should be used out of the gate yet or if there should be more thought before it gets utilized. Wouldn't mind any input on this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ required-features = ["render"]
 bevy = { version = "0.12", default-features = false, features = [
     "bevy_asset",
 ] }
-egui = { version = "0.24.0", default-features = false, features = ["bytemuck"] }
+egui = { version = "0.25.0", default-features = false, features = ["bytemuck"] }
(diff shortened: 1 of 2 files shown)


MickHarrigan:
Converted this to a draft as with some further digging the PRs in both [`bevy_input`]() and [`egui`]() that updated the winit crate to 0.29 have a bit more nuance to them that this change should take note of. I will continue to read through their changes and see what is the best course of action soon.

Vrixyz:
Now with https://github.com/bevyengine/bevy/pull/11400 ; from `KeyboardInput`, we can now use `key_code` for physical key, and `logical_key` for logical key.

xangelix:
https://github.com/emilk/egui/releases/tag/0.26.0 is also now available.

mvlabat:
Egui 0.26 is now in `main`, thanks for the PR!

PR #251, opened 2024-01-20
Make fields inside WindowSize pub

BeastLe9enD:


mvlabat:
Oh, that was an oversight. Thanks for the PR!

PR #256, opened 2024-02-06
Fix @ on AZERTY on wasm

Vrixyz:
Another attempt to https://github.com/mvlabat/bevy_egui/pull/181

This PR retrieves the user agent for better platform detection.

Not much tested, it seemed on wasm I was receiving physical keys, so my next step is to check how it's going with bevy 0.13


frewsxcv:
https://bevyengine.org/news/bevy-0-13/#winit-upgrade

mvlabat:
I made some changes to the code, now this should be a bit cleaner and also work for iphones with connected keyboards (yeah, I decided to test that, lol). Huge thanks for moving all the input story to completion, it works wonderfully now!

PR #262, opened 2024-02-19
Update Arboard dep

naomijub:


mvlabat:
Hi, could you clarify why you want the dependency updated in the library? If you want to use a newer version of `arboard` in your app, having an older version specified here shouldn't prevent you from using the newer one, since they are semver compatible. (I.e. having `arboard 3.2.0` and `arboard 3.3.1` in your `Cargo.toml` will resolve in `arboard 3.3.1`.)

But if we raise the version in the library, that means that we are forcing our users to update. Which isn't a necessarily bad thing, but who knows.. I probably wouldn't raise it without a need. So unless there's a vulnerability fix or a feature that we need for example, I'd rather leave it as is.

naomijub:
it was mostly for a personal thing, I can keep it updated on my fork

mvlabat:
@naomijub yeah, you can just update it with `cargo update`, and you'll see that it'll install a newer version in Cargo.lock.

PR #265, opened 2024-02-27
Fix text editing keybinds

eero-lehtinen:
Copy, paste, cut, undo, redo and maybe others were bound to WIN instead of CTRL outside of macOS, which was wrong. It broke after updating to the latest Bevy. This pr binds them back to CTRL.

I also have to prevent text input when win is pressed because the command check doesn't work by itself anymore.

I don't have a mac available right now, but I assume it was broken for macOS too, because copy is CMD-C on macOS. The egui docs also tell you to use the same thing for command as mac_cmd.

### DIFF
--- a/src/systems.rs
+++ b/src/systems.rs
@@ -163,7 +163,7 @@ pub fn process_input_system(
         win,
     } = *input_resources.modifier_keys_state;
     let mac_cmd = if *context_params.is_macos { win } else { false };
-    let command = if !*context_params.is_macos { win } else { ctrl };
+    let command = if *context_params.is_macos { win } else { ctrl };
 
     let modifiers = egui::Modifiers {
         alt,
@@ -266,7 +266,7 @@ pub fn process_input_system(
         }
     }
 
-    if !command || !*context_params.is_macos && ctrl && alt {
+    if !command && !win || !*context_params.is_macos && ctrl && alt {
         for event in input_events.ev_received_character.read() {
             if event.char.matches(char::is_control).count() == 0 {
                 let mut context = context_params.contexts.get_mut(event.window).unwrap();


src/systems.rs
```diff
@@ -163,7 +163,7 @@ pub fn process_input_system(
         win,
     } = *input_resources.modifier_keys_state;
     let mac_cmd = if *context_params.is_macos { win } else { false };
-    let command = if !*context_params.is_macos { win } else { ctrl };
+    let command = if *context_params.is_macos { win } else { ctrl };
```
Vrixyz, line 166: yes, that fix seems correct..

I really wish we could make some tests on that :o ; it's a pain to verify

src/systems.rs
```diff
@@ -266,7 +266,7 @@ pub fn process_input_system(
         }
     }
 
-    if !command || !*context_params.is_macos && ctrl && alt {
+    if !command && !win || !*context_params.is_macos && ctrl && alt {
```
mvlabat, line 269:
```suggestion
    if !command || !*context_params.is_macos && ctrl && alt {
```
I think adding `&& !win` to the condition makes it so that we no longer ignore character input on such hotkeys as `Ctrl + C` on Windows (i.e. on pressing `Ctrl + C` it'll both copy and paste the `c` character)

eero-lehtinen: The logic is a bit weird because it's negated, but it definitely works correctly as I tested it on Linux and it works the same on Windows.

It checks that both `Win` and `Ctrl` are not pressed before typing characters. I wouldn't want to type anything if pressing e.g `Win+Tab` or `Ctrl+C` .

mvlabat: Yeah, my bad, I for some reason confused `win` (a modifier key) with the Windows platform. I surely haven't had enough sleep today 😄

mvlabat:
Thank you for the PR!

PR #270, opened 2024-03-27
Better error on Web when missing web_sys_unstable_apis

Vrixyz:
<details><summary>Before:</summary>
<p>

image

</p>
</details> 

<details><summary>After:</summary>
<p>

image

</p>
</details>

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -50,6 +50,13 @@
 //!
 //! - [`bevy-inspector-egui`](https://github.com/jakobhellermann/bevy-inspector-egui)
 
+#[cfg(all(
+    feature = "manage_clipboard",
+    target_arch = "wasm32",
+    not(web_sys_unstable_apis)
+))]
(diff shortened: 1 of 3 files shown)


.cargo/config.toml
```diff
@@ -1,6 +1,9 @@
 [alias]
 run-wasm = ["run", "--release", "--package", "run-wasm", "--"]
 
+[build]
+target = "wasm32-unknown-unknown"
+
 # Using unstable APIs is required for writing to clipboard
 [target.wasm32-unknown-unknown]
-rustflags = ["--cfg=web_sys_unstable_apis"]
+# rustflags = ["--cfg=web_sys_unstable_apis"]
```
Vrixyz, line 9: 😅

Vrixyz:
Not 100% sure if include_str! is wokring as-is when publishing, or if it's good practice to add a folder for the errors, but I think it looks better than the alternative which doubles the error message in a less readable fashion:

image

PR #271, opened 2024-03-28
update to egui 0.27

jakobhellermann:
Pretty uneventful update, here's a list to the breaking changes [egui/CHANGELOG.md](https://github.com/emilk/egui/blob/master/CHANGELOG.md#%EF%B8%8F-breaking)

> Response::clicked* and Response::dragged* may lock the Context, so don't call it from a Context-locking closure.

I don't think this is done in `bevy_egui` so that should be fine.


mvlabat:
Cheers! I'll try to release it soon together with #270

Issue #176, opened 2023-04-14
Reactive mode support


thmxv:
The egui demo has something called "reactive" mode that does not redraw continuously, usually implemented with `request_repaint_after()`. And bevy allow to use a custom main loop with `set_runner()`. However I fail to understand how to tie both of those features together. I would like to render, both the UI and the 3D viewport, only after an event.

thmxv:
I discovered that there is no need to write a custom main loop. Bevy has already a setting to work in "reactive mode" like egui `App::new().insert_resource(WinitSettings::desktop_app())` and the two works perfectly in conjunction.

A-Totally-Normal-Name:
> .insert_resource(WinitSettings::desktop_app())

Doesn't this make the entire app run in reactive mode? Some may only want the egui menus to run in reactive mode instead of their whole bevy app. Hence, this solution isn't necessarily the best unless I'm missing something.

thmxv:
Yes, it makes both bevy and egui run in reactive mode, but this is what I need.

PR #278, opened 2024-04-23
Request Redraw only if really needed. Fixes #277

### LINKED ISSUE #277: How can I stop bevy_egui from sending RequestRedraw whenever I move the mouse?
Hello ! 👋 

bevy_egui is sending a `RequestRedraw` event to Bevy in the `process_output_system`, here:

https://github.com/mvlabat/bevy_egui/blob/6021f22a0e69ed802844a259d4b43d94007027ec/src/systems.rs#L535-L537

... when any of the EGUI contexts need a repaint

I'm using egui only for debugging purposes (a debug Window, similar to the `bevy_inspector_egui`) but I have a problem: even when my EGUI window is not opened at all (so, I'm not even running the systems to draw it), the EGUI systems run and send the `RequestRedraw` event

This is a problem for me because I want to control the frame rate, and I do not want to run a new frame, even if I receive an input event from a device or the window

**Repro steps** are fairly easy: in the `simple.rs` example of this repo, comment out the system. It will only show a black screen, but if you put a `println!` log in the piece of code sending the `RequestRedraw` event linked above, it will be called **whenever the cursor moves on top of the screen**

How can I disable this behaviour from my app? I'm trying to remove the EGUI Context from the PrimaryWindow, but there is an unwrap inside of `bevy_egui` that is expecting it, so it panics.

Since I think this cannot be done from outside of `bevy_winit`, I'm trying to modify the source code to run this system only when there is anything actually being painted, but I'm still digging through the code to learn how I can do it, so any advice is welcome 🙂

Maximetinu:
Hello,

I had this problem that I described at this issue:

- #277

where it is a blocker for my use case that `bevy_egui` sends `RequestRedraw` events, even when I'm not painting any EGUI window.

So, this PR fixes the `process_output_system` so that it doesn't emit that event when nothing is being painted.

It will keep working as usual when EGUI is painting.

----

On another note:

I'm not sure that `RequestRedraw` is really needed. If the `UpdateMode` of the Bevy's `WinitSettings` resource is `UpdateMode::Continuous`, Bevy will redraw anyway. If it is `UpdateMode::Reactive` or `UpdateMode::ReactiveLowPower`, both of them should answer to Window events, like moving the cursor ([see here in bevy_winit](https://github.com/bevyengine/bevy/blob/c593ee1055047a64501efa5de5885c9d85547af3/crates/bevy_winit/src/lib.rs#L387-L404))

However, I'm not removing it in this PR to be cautious and not reintroduce previous bugs like 
- #240
- #239
- #176

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -288,6 +288,13 @@ pub struct EguiRenderOutput {
     pub textures_delta: egui::TexturesDelta,
 }
 
+impl EguiRenderOutput {
+    /// Returns `true` if the output has no Egui shapes and no textures delta
+    pub fn is_empty(&self) -> bool {
+        self.paint_jobs.is_empty() && self.textures_delta.is_empty()
+    }
+}
+
 /// Is used for storing Egui output.
 #[derive(Component, Clone, Default)]
 pub struct EguiOutput {
--- a/src/systems.rs
+++ b/src/systems.rs
@@ -532,7 +532,10 @@ pub fn process_output_system(
         #[cfg(not(windows))]
         set_icon();
 
-        if ctx.has_requested_repaint() {
+        let needs_repaint = !context.render_output.paint_jobs.is_empty()
+            && !context.render_output.textures_delta.is_empty();
+
+        if ctx.has_requested_repaint() && needs_repaint {
             event.send(RequestRedraw);
         }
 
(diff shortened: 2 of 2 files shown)


mvlabat:
Hi, thank you for the PR! Lgtm, apologies for taking that long to review it

Issue #29, opened 2021-05-21
Text input doesn't bring up virtual keyboard on mobile


stmdhms:
This was recently fixed in egui itself (https://github.com/emilk/egui/pull/247), but doesn't currently work correctly in bevy_egui.

OfficialURL:
This would be solved by #27.

mvlabat:
I'm afraid just upgrading to 0.12 doesn't fix it. I guess we need to port the changes that were done in https://github.com/emilk/egui/pull/253 do `bevy_egui`. Unfortunately, I'm not able to look into those atm myself, but I'm happy to accept a PR if anyone wants to tackle that.

miketwenty1:
Is it still possible to bring up the mobile keyboard with a WASM build?

mvlabat:
Still impossible, unfortunately. If anyone wants to make a PR for this, I'll gladly accept it.

oscrim:
Hello, Ive been working on this on [this branch](https://github.com/oscrim/bevy_egui/tree/touch-and-text-input) (which also containts the work from #180 ). Due to winit lacking support for composition events in the web (as can be seen [here](https://github.com/rust-windowing/winit/issues/1497)) I excluded the event listeners for them as it currently sends them all out at once when the composition ends which causes unwanted behavior.

v-kat:
@oscrim did your code fix mobile keyboard issues with bevy_egui? I wasn't sure if this was an IME issue that has to be explored at a lower level.

oscrim:
It was more of a work around than a proper fix. It's been a while since I looked at it

v-kat:
Do you think it's something you could eventually upstream? I was still trying to understand if there was a bug in IME somewhere or if some of the code from egui needed to be adapted/bridged to bevy. (Maybe from https://github.com/emilk/egui/blob/master/crates/eframe/src/web/text_agent.rs )

v-kat:
I might just steal your code and maybe try to upstream it >.>

mvlabat:
Implemented in bevy_egui 0.30

PR #279, opened 2024-04-26
Mobile web virtual keyboard support

v-kat:
Fixes https://github.com/mvlabat/bevy_egui/issues/29

Mostly stolen code from https://github.com/emilk/egui/blob/master/crates/eframe/src/web/text_agent.rs and https://github.com/oscrim/bevy_egui/tree/gui-touch-event still has the same issue as https://github.com/emilk/egui/issues/4403 but seems to be mostly working. I had to use `once_cell` because iOS safari doesn't let you set an input focus unless the code is directly executed in the input callback. I also didn't proxy all of the key press events and am only doing backspace due to another iOS safari issue but could change that later.... Seems to work about as well as the egui mobile web virtual keyboard in eframe but there might be some issues.

### DIFF
--- a/src/text_agent.rs
+++ b/src/text_agent.rs
@@ -150,10 +150,10 @@ pub fn install_text_agent(
     if let Some(true) = is_mobile() {
         let input_clone = input.clone();
         let sender_clone = sender.clone();
-        let closure = Closure::wrap(Box::new(move |_event: web_sys::InputEvent| {
+        let closure = Closure::wrap(Box::new(move |event: web_sys::InputEvent| {
             let text = input_clone.value();
 
-            if !text.is_empty() {
+            if !text.is_empty() && !event.is_composing() {
                 input_clone.set_value("");
                 if text.len() == 1 {
                     input_clone.blur().ok();


v-kat:
I might need to clean up some of the logic a tiny bit to be more sensical. eframe was doing weird things with canvas resizing that didn't seem strictly necessary but I also didn't have a lot of different kinds of test devices. It seems to work as is and I'm not sure when I'll have the effort to really clean it up (the egui textEdit interaction and highlighting on mobile are still not very good as a base too). This also followed the weird eframe approach where a double tap is needed to open up the keyboard. I might try to fix that later too.

v-kat:
@mvlabat any chance of a review?

mvlabat:
@v-kat hi, thank you for the PR! And apologies for neglecting it for so long. I've just tested it on my Android: for some reason, the keyboard disappears immediately after opening it. It seems like there's another click/touch event that follows immediately after that causes this behaviour.

src/text_agent.rs
```diff
@@ -0,0 +1,245 @@
...
+                    key: egui::Key::Backspace,
+                    physical_key: None,
+                    pressed: true,
+                    modifiers: egui::Modifiers::NONE,
+                    repeat: false,
+                });
+            }
+        }) as Box<dyn FnMut(_)>);
+        document.add_event_listener_with_callback("keydown", closure.as_ref().unchecked_ref())?;
+        closure.forget();
```
mvlabat, line 115: Instead of forgetting the closure, can we register it in the `SubscribedEvents` resource? (Feel free to move it to `lib.rs` as it resides in the `web_clipboard` module atm.)
This way users will be able to unsubscribe manually in case they'll want to have some sort of multi-app setup on their page.

src/text_agent.rs
```diff
@@ -0,0 +1,245 @@
...
+static AGENT_ID: &str = "egui_text_agent";
+
+#[allow(missing_docs)]
+#[derive(Clone, Copy, Debug, Default)]
+pub struct VirtualTouchInfo {
+    pub editing_text: bool,
+}
+
+pub static VIRTUAL_KEYBOARD_GLOBAL: Lazy<Mutex<VirtualTouchInfo>> =
+    Lazy::new(|| Mutex::new(VirtualTouchInfo::default()));
```
mvlabat, line 26: Can we use stdlib's [`OnceCell`](https://doc.rust-lang.org/std/cell/struct.OnceCell.html) instead?

v-kat:
> @v-kat hi, thank you for the PR! And apologies for neglecting it for so long. I've just tested it on my Android: for some reason, the keyboard disappears immediately after opening it. It seems like there's another click/touch event that follows immediately after that causes this behaviour.

Are you seeing the same behavior on https://vpetmon.com I think there's maybe a conflict with the copy paste feature and duplicate events but need to double check

src/text_agent.rs
v-kat: The value gets set multiple times so I don't think so.

v-kat:
There seems to be some bugs after I merged in 0.14 I'll investigate and try to update later.

src/text_agent.rs
mvlabat: It can contain `Mutex` the same way `Lazy` does rn.

v-kat: I think the std `OnceCell` doesn't let you implement `Sync` and the std `OnceLock` only lets you do `static` so this is actually necessary despite the lack of true threads.

v-kat:
I tested it and seemed to be working on 0.14 still looking at using `SubscribedEvents` working through some minor type annoyances with the `EventClosure` closure. There's some jank with the TextEdit cursor not always easily appearing on touch and then the keyboard input not doing anything until the input actually has focus with the cursor.

src/text_agent.rs
v-kat: Addressed in https://github.com/mvlabat/bevy_egui/pull/279/commits/5985a6a596b258f36f442cd8a1201a17014ce6a3 I did multiple different `SubscribedEvents` due to some annoyances with types but could switch things.

v-kat:
Maybe https://github.com/emilk/egui/issues/4500 will eventually fix some of the issues. Otherwise maybe this is ready for review again.

v-kat:
https://blog.rust-lang.org/2024/07/25/Rust-1.80.0.html has LackLock but I didn't want to try it. The TextEdit cursor focus issues on egui are pretty annoying and I might look at https://github.com/emilk/egui/issues/4500 later to make this way less janky.

mvlabat:
I've just tested it on both Android and iOS, and I'm seeing the same issue: the keyboard closes immediately after opening it

v-kat:
> I've just tested it on both Android and iOS, and I'm seeing the same issue: the keyboard closes immediately after opening it

Can you tell me how you tested it and if it's the same on https://vpetmon.com I have at least 2 android users and an iOS user that are using it currently although the focus is still buggy

mvlabat:
I test your PR with running the `ui` example in web: `cargo run-wasm --example ui --host 0.0.0.0`.

The focus behaviour is indeed consistent with what you're saying, it does require 2 taps to bring the keyboard up. Yet in the `ui` example the keyboard appears and is closed immediately on the second tap.

I just tested https://vpetmon.com/ on my Android (Google Chrome browser): the keyboard is indeed brought up, it doesn't disappear immediately like it does in the `ui` example, but no text appears in the input when I press keyboard buttons.

v-kat:
Apparently `prevent_default_event_handling: false,` set on the `WindowPlugin` causes the keyboard not to work at least in your example. The other issue you're describing is the bad focus behavior with egui TextEdits that's getting fixed/examined upstream. If you click until you see a cursor or flashing bar in the text field the keyboard should respond. I'm not sure if having this working without `prevent_default_event_handling` is possible but I can look more. I'm still waiting for upstream changes to improve the focus behavior. I don't see much else I can do otherwise. I can add a note about `prevent_default_event_handling` to the readme?

mvlabat:
Yeah, I can confirm that setting the flag back to true works around the focus issue. Unfortunately, I still can't get any text entered in the input field. 

It works the same way on vpetmon for me.

https://github.com/user-attachments/assets/6c7fd217-fc6c-4191-8220-65f7a46bc71f

v-kat:
> Yeah, I can confirm that setting the flag back to true works around the focus issue. Unfortunately, I still can't get any text entered in the input field.
> 
> It works the same way on vpetmon for me (I can't get any text entered there either).
> 
>  Screen_Recording_20240804_004445_Chrome.mp4

You have to click multiple times until there's a blinking cursor indicating the TextEdit can receive input 
Screenshot_2024-08-03-14-59-53-82_3aea4af51f236e4932235fdada7d1643
I need to tweak the logic a tiny bit to not do spurious opens but getting always correct focus is an open egui issue.

v-kat:
lame_example
It's not very good and kind of hard to do successfully given the small sizes of TextEdits but does work.... https://github.com/emilk/egui/pull/4855 will hopefully help and I'll also steal any other changes. LMK if you want a different `prevent_default_event_handling` fix. I kind of want to put this down and/or just get it merged despite the jank but am fine if you want to wait or something else.

v-kat:
Using `egui = { git = "https://github.com/micmonay/egui", default-features = false, branch = "fix-keyboard", features = ["bytemuck"] }` for better TextEdit focus handling. Might maybe steal the PointerEvent change in https://github.com/emilk/egui/pull/4848 but for now android always works on 1 touch for the input (as long as using https://github.com/emilk/egui/pull/4855) and safari requires a double click due to ordering with the `process_input_system` firing after the iOS native safari event that's only needed due to browser quirks (I couldn't think of a better way to fix this). Maybe still some other issues but seems pretty good when using the `crates/egui/src/widgets/text_edit/builder.rs` change in 4855

opfromthestart:
Good work on this, this branch works for the most part. However, when I click into the text box, the page zoom and scroll is reset, which means that I cannot double click on it to start typing. How could I help to fix this?

v-kat:
> Good work on this, this branch works for the most part. However, when I click into the text box, the page zoom and scroll is reset, which means that I cannot double click on it to start typing. How could I help to fix this?

Are you using the egui branch I mentioned and Android and the same example testbed?

opfromthestart:
I am not using the same testbed, as I am using it in my own game. On modifying the egui branch, do you mean putting it in the bevy_egui Cargo.toml? That made it take one click instead of 2, but it still repositions the website on every key input.

v-kat:
I haven't seen any zoom or canvas behavior like you're describing on Android with the UI example.

I'm mostly finished with this so please close it if it's a big issue. I'm waiting for egui upstream changes and no longer working on this.

v-kat:
> I am not using the same testbed, as I am using it in my own game. On modifying the egui branch, do you mean putting it in the bevy_egui Cargo.toml? That made it take one click instead of 2, but it still repositions the website on every key input. Also, when the canvas is maximized/made fullscreen, the keyboard no longer appears when clicking on a text input. I cannot test if these are due to egui as the egui demo is in full screen and repositions elements to be on-screen when the keyboard appears.

If you give me more debug info I can try to help. I'm using it in prod on https://vpetmon.com with the egui branch I referenced and it seems okay with:

```rust
DefaultPlugins
                .set(WindowPlugin {
                    primary_window: Some(Window {
                        title: "Games VPetmon".to_string(),
                        fit_canvas_to_parent: true,
                        ime_enabled: true,
                        // prevent_default_event_handling: false, // to enable copy paste
                        // mode: WindowMode::BorderlessFullscreen,
                        ..default()
                    }),
                    ..default()
                })
                .set(AssetPlugin {
                    meta_check: AssetMetaCheck::Never,
                    ..default()
                }),
```

opfromthestart:
The main change I have is that the canvas does not take up the entire screen, while the given example does. An example is at 139.144.52.74:12345, it may be a while until I can update it though.

v-kat:
@mvlabat the tests looked like they are failing with a build error. I'm still ideally waiting for a PR to be merged https://github.com/emilk/egui/pull/4855 so that this works better but think it works decently. 

I didn't see a good way to fix 2 touches being needed to focus on an input on iOS. Even when trying to thread through the context it would need to be accessed repeatedly in the `touchend` handler due to iOS security on setting focus which brought up `'static` lifetime issues.

I'd like to see this merged soon and dunno how long the egui PR and version update will take to happen.

Hopefully it's good enough.

mvlabat:
Yeah I was even thinking about merging and releasing it yesterday, considering that it'll likely work correctly once the linked PR is merged into Egui itself. But I decided to wait for a new Egui version anyway, just to be able test it with the final code that lands into Egui and be more confident that it works as we want.

And thank you for keeping this updated! I do hope we merge it before the next release of beyv_egui :)

v-kat:
Thanks. I found out a hacky way to do 1 touch on iOS too and will hopefully update this soon with a better version (I didn't see a way to solve 'static context lifetime issues even with the immutable feature).

src/text_agent.rs
v-kat: I removed it and added a minimum rust version to use std sync `LazyLock` I think I need a 'static....

mvlabat:
Already doing that :) I'll definitely merge it before the release, might push some changes to the branch as well

mvlabat:
I just merged to main. Thank you again!

I had to finish it in a separate branch and then, when I squashed it, I realised it only attributed me as an author. I tried to trick github into linking the commit to your profile as well but it wouldn't work because of the fake email (that's why you see all these force-pushes to `main` :D)

mvlabat:
Ok, I just found the way lol. Butchering the repo history was totally worth it 😄

bananaturtlesandwich:
hi i've got a website up at https://skyweb.spuds.casa which this isn't working for - any idea why? the source is at https://github.com/bananaturtlesandwich/skyweb with the ui code in config.rs

PR #282, opened 2024-05-23
fix light in `render_to_image_widget` example

rlidwka:
Issue https://github.com/bevyengine/bevy/issues/3462 has been fixed. So since Bevy 0.13, render lights can be specified per RenderLayer.

So our example needs to be updated, because right now render pass does not have any light at all.

I've decided to reuse the same light by specifying `RenderLayers::all()`, although adding a separate one (with preview_pass_layer and without) is also fine.


mvlabat:
Good catch, thank you

Issue #296, opened 2024-07-18
Input State of Mod Keys get's stuck


RobWalt:
How to reproduce:

- Adjust the `simple.rs` example

```rust
use bevy::prelude::*;
use bevy_egui::{EguiContexts, EguiPlugin};

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_plugins(EguiPlugin)
        // Systems that create Egui widgets should be run during the `CoreSet::Update` set,
        // or after the `EguiSet::BeginFrame` system (which belongs to the `CoreSet::PreUpdate` set).
        .add_systems(Update, ui_example_system)
        .run();
}

fn ui_example_system(mut contexts: EguiContexts, mut text: Local<String>) {
    egui::Window::new("Hello").show(contexts.ctx_mut(), |ui| {
        ui.label("world");
        ui.text_edit_singleline(&mut *text);
    });
}
```

- Run the example
- Write something into the text edit field "asdfg"
- switch focus of windows with any modifier keys (for me it's SuperLeft+Arrow, we have apple users reporting this as well though)
- click with the mouse into the window to gain back focus
- try writing in the text edit
- This won't work since the mod key state is stuck and the textfield thinks you're inputting "Mod+Letter" instead of just "Letter"

RobWalt:
I temporarily changed the bevy dependency to point to the fix here https://github.com/bevyengine/bevy/pull/14379 and it seems like this also solves this issue in a more detailed way. 

I'll close this issue here then and try to help in the bevy repo to get it going.

PR #297, opened 2024-07-18
Simple fix to clear wrong input state

RobWalt:
Kind of solves #296

This fix is pretty stupid though. Maybe there's something better we can do?

*edit:

I tried to make the reset happen not as often and only when we lose focus of <ALL> windows. I hope that this will rule out any unexpected state resets.

### DIFF
--- a/examples/simple.rs
+++ b/examples/simple.rs
@@ -11,8 +11,9 @@ fn main() {
         .run();
 }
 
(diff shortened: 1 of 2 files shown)


RobWalt:
Closing because there's an alternative solution directly in bevy. See comments in #296

Issue #232, opened 2023-11-27
Sticky shift key


doonv:
## The bug

Pressing the <kbd>Shift</kbd> key once causes it to remain pressed after releasing it. This usually doesn't affect anything but it does effect `ScrollArea` scrolling, which forces horizontal scrolling after <kbd>Shift</kbd> was pressed.

## Example

```rs
use bevy::prelude::*;
use bevy_egui::{egui, EguiContexts, EguiPlugin};

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_plugins(EguiPlugin)
        // Systems that create Egui widgets should be run during the `CoreSet::Update` set,
        // or after the `EguiSet::BeginFrame` system (which belongs to the `CoreSet::PreUpdate` set).
        .add_systems(Update, ui_example_system)
        .run();
}

fn ui_example_system(mut contexts: EguiContexts) {
    egui::Window::new("Hello").show(contexts.ctx_mut(), |ui| {
        let pressed = ui.input(|i| i.modifiers.shift);
        info!("pressed? {pressed}");
        ui.label("world");
    });
}
```

It will start out printing `pressed? false`. But after tapping <kbd>Shift</kbd> it will only print `pressed? true`

doonv:
Wait no it looks like it's a bevy problem?

doonv:
Yeah it's a bevy problem. Should be fixed by https://github.com/bevyengine/bevy/pull/10702.

SludgePhD:
I opened a PR with a fix in https://github.com/mvlabat/bevy_egui/pull/298

PR #298, opened 2024-07-19
Clear modifier state when focus is lost

SludgePhD:
This does the same thing bevy does for the `ButtonInput<KeyCode>` resource here: https://github.com/bevyengine/bevy/blob/d8d49fdd1393ab543a0ed66da9a255fda7d66572/crates/bevy_input/src/keyboard.rs#L151-L155

Fixes https://github.com/mvlabat/bevy_egui/issues/232

I'm not sure why all the fields in `InputEvents` are public, since it makes this a breaking change and isn't required for things to compile.

### DIFF
--- a/src/systems.rs
+++ b/src/systems.rs
@@ -8,7 +8,7 @@ use bevy::{
         system::{Local, Res, SystemParam},
     },
     input::{
-        keyboard::{Key, KeyCode, KeyboardInput},
+        keyboard::{Key, KeyCode, KeyboardFocusLost, KeyboardInput},
         mouse::{MouseButton, MouseButtonInput, MouseScrollUnit, MouseWheel},
         touch::TouchInput,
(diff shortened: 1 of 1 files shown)


mvlabat:
Awesome, thank you for the fix!

mvlabat:
Btw, the motivation for public fields was to allow people to re-use parts of `bevy_egui` logic if they want to re-implement some parts of input handling. Since we are on 0.x, I'm fine with breaking changes, as I usually publish a new version on Egui or Bevy release, which tends to introduce a lot of breaking changes anyway.

If we ever reach 1.0, the fields will become private. Or it might as well happen sooner, for example, after we merge #299 and provide some API for people to handle tricky input scenarios such as worldspace UI, etc.

Anyway, public fields of resources is something I'm open to reconsider.

PR #299, opened 2024-07-24
implement worldspace ui

Schmarni-Dev:
this reimplements #214 while referencing a lot of parts of the pr

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,7 @@ bevy = { version = "0.14.0", default-features = false, features = [
 egui = { version = "0.28", default-features = false, features = ["bytemuck"] }
 bytemuck = "1"
 webbrowser = { version = "1.0.1", optional = true }
+wgpu-types = "0.20"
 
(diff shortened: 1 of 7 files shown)


mvlabat:
Hi, thank you a lot for the PR, this is great! I'm still trying to find the time+energy to review and test it, so apologies for the delay.

I'm thinking how we should handle inputs for worldspace ui. I guess we need some sort of API to allow users pass mouse events casted to 3D surfaces.. Maybe users can manually emit native Bevy events by mapping keyboard/mouse events if they detect a cursor is over a 3D surface.

I think it's something we can try exploring in examples before providing a better API for it (it doesn't necessarily have to be introduced in this PR).

Schmarni-Dev:
i already have code to interact with world space egui using bevy_mod_picking here https://github.com/NexusSocial/nexus-vr/tree/main/crates/egui-picking, i just need to update it to use this impl and add improvements. i would like to upstream that to here once bevy upstreams bevy_mod_picking in 0.15

mvlabat:
Merged as part of #304. Thank you for your help to drive this to completion!

Issue #300, opened 2024-08-13
Clipboard feature will not compile, web-sys API change


xangelix:
The `web-sys` clipboard API is now a bit different, see:

https://github.com/rustwasm/wasm-bindgen/pull/3992

```
error[E0308]: mismatched types
   --> /home/tux/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bevy_egui-0.28.0/src/web_clipboard.rs:233:13
    |
233 |         let Some(clipboard) = nav.clipboard() else {
    |             ^^^^^^^^^^^^^^^   --------------- this expression has type `Clipboard`
    |             |
    |             expected `Clipboard`, found `Option<_>`
    |
    = note: expected struct `Clipboard`
                 found enum `std::option::Option<_>`

For more information about this error, try `rustc --explain E0308`.
```

rapushka:
does anyone know a workaround for this?

i'm using [this template](https://github.com/bevyengine/bevy_github_ci_template) for my project and i'm unable to build wasm due to this error, but i only need egui for debug builds! i use it as a dependency for [bevy editor pls](https://github.com/jakobhellermann/bevy_editor_pls), and so i don't need it in my wasm builds at all!

```toml
[dependencies]
bevy_editor_pls = "0.9"
```

i tried to put this one in `[dev-dependencies]` but it doesn't work, cuz as far as i understand it can't work with binary projects

mvlabat:
Thank you for the report! The fix is released in the 0.29 version.

PR #301, opened 2024-08-15
Adapt to `web-sys` clipboard api change

no-materials:
Addresses #300


mvlabat:
Thank you for the fix! I'll release a new version with it today.

Issue #294, opened 2024-07-08
Support epaint::Shape::Callback(PaintCallback)


PPakalns:
https://github.com/mvlabat/bevy_egui/blob/6705757bc29f7194e58129276100a4d52b93aadc/src/egui_node.rs#L222

PaintCallback is not supported. This callback allows painting 3D views, visualizations, etc inside egui Windows using underlying backend (bevy rendering / wgpu). 

Maybe PaintCallback could somehow get expanded into render graph subgraph.
https://docs.rs/bevy_render/latest/bevy_render/render_graph/struct.RenderGraph.html#method.add_sub_graph

mvlabat:
I'm happy to accept a PR implementing this. Unfortunately, I don't have enough capacity to work on this myself.

PR #303, opened 2024-08-17
Paint callback support

PPakalns:
Resolves https://github.com/mvlabat/bevy_egui/issues/294

Bevy render graph nodes create render passes, paint callback can not be expanded into a render graph subgraph.

Graph node can contain custom rendering code or, for example, [Render Phase](https://docs.rs/bevy_render/latest/bevy_render/render_phase/index.html) abstraction to render multiple type of resources with multiple render pipelines in the current render pass.
This implementation supports bevy rendering logic with custom render implementation.

For now native wgpu RenderPass can not be used because of the following issue: https://github.com/bevyengine/bevy/issues/13225

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,9 @@ render = ["bevy/bevy_render"]
 serde = ["egui/serde"]
 
 [[example]]
+name = "paint_callback"
+required-features = ["render"]
+[[example]]
 name = "render_to_image_widget"
 required-features = ["render"]
 [[example]]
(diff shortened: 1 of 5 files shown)


mvlabat:
Awesome work, thank you! I'm also amazed how it _just works_ in browsers

mvlabat:
I pushed a couple of fixes to make CI happy and also did a small and a bit of unnecessary refactor, I hope it doesn't break anything :D

mvlabat:
I've just found a copy-paste error that I added: will be fixed in #304 if I manage to implement callbacks support for worldspace UI based on the current implementation (or separately).

PR #306, opened 2024-08-27
Paint callback functionality improvements

PPakalns:
Support executing additional render passes or adding command buffers from paint callback

### DIFF
--- a/src/egui_node.rs
+++ b/src/egui_node.rs
@@ -382,11 +382,11 @@ impl Node for EguiNode {
         }
     }
 
-    fn run(
+    fn run<'w>(
         &self,
         _graph: &mut RenderGraphContext,
-        render_context: &mut RenderContext,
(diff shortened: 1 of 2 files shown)


mvlabat:
Hi, thanks for the PR!

Recently, we've also added this node: https://github.com/mvlabat/bevy_egui/blob/3aed12f9292d46c81a1cde1160198c6fb572e253/src/egui_render_to_texture_node.rs

It duplicates probably ~80% of `egui_node.rs` code, which I'm not a fan of. But until we refactor that, we'll have to introduce any changes to both of them.

PPakalns:
Logic added, feel free to edit branch if minor fixes are needed.

PPakalns:
Ping in case of missing this in github feed :)

mvlabat:
Looks great, thank you!

PR #309, opened 2024-09-24
Update to bevy 0.15

Vrixyz:
Currently testing on bevy 0.15 rc3.

- cursor handling changed, we rely on the new feature `custom_cursor`.
- more info about retained rendering world update: https://github.com/Vrixyz/bevy_egui/pull/5

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -79,10 +79,14 @@ mod text_agent;
 ))]
 pub mod web_clipboard;
 
-use bevy_winit::cursor::CursorIcon;
 pub use egui;
 
 use crate::systems::*;
+#[cfg(target_arch = "wasm32")]
+use crate::text_agent::{
+    install_text_agent, is_mobile_safari, process_safari_virtual_keyboard, propagate_text,
+    SafariVirtualKeyboardHack, TextAgentChannel, VirtualTouchInfo,
+};
 #[cfg(feature = "render")]
 use crate::{
     egui_node::{EguiPipeline, EGUI_SHADER_HANDLE},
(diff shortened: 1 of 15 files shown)


src/lib.rs
```diff
@@ -797,6 +798,8 @@ pub struct EguiContextQuery {
     pub render_target_size: &'static mut RenderTargetSize,
     /// [`Window`] component, when rendering to a window.
     pub window: Option<&'static mut Window>,
+    /// Cursor for the
```
Friz64, line 801: Cursor for, for the, the Cursor for, Cursor the, the for, Cursor the Cursor for the

Vrixyz: Thanks for pointing that out!

It's actually a non trivial change: The handling of the cursor is changed in bevy 0.15 ; in this PR I moved the initialization of the cursor to the user through `configure_cursor` :
https://github.com/mvlabat/bevy_egui/blob/3be113b82f8b7122b33f92467d20d2560e0d7722/examples/ui.rs#L56-L60

But I think this code snippet should live with bevy_egui (or bevy_something :thinking: ; but in the meantime we should be able to add it if not present.)

PPakalns:
> Currently testing on [bevy commit 9386bd0](https://github.com/bevyengine/bevy/commit/9386bd0114c44c9f00a2e9c41db1225aaa78d159)
> 
>     * cursor handling changed, I think we should initialize it for all windows if not present (currently not implemented), see discussion: https://github.com/mvlabat/bevy_egui/pull/309/files#r1780569354

Just brainstorming cursor handling.
Maybe we could add for the window "EguiCursorIcon" component that bevy_egui sets.

We can provide a system that is added by default that correctly sets CursorIcon for the window from the egui EguiCursorIcon value.

But user could opt out:
User can handle manually EguiCursorIcon, it could be useful in cases where user would like to read the EguiCursorIcon state, but would want to have custom logic  (For example user is utilizing a lot of custom cursor icons. He needs information from egui about cursor type, but wants to manually set their own cursor icons from icon set that do not map to SystemCursorIcon enum.

In this case it would be nice if egui would not set CursorIcon component directly.

---

Probably this functionality could be added in a PR at later date.

Vrixyz:
> bevy pushed version 0.15.0-rc.1

- Thanks! I started to work on that at https://github.com/Vrixyz/bevy_egui/pull/5
- But the migration is not trivial due to https://github.com/bevyengine/bevy/pull/15320/files

Friz64:
Also relevant: https://github.com/bevyengine/bevy/pull/16093

Vrixyz:
> Also relevant: [bevyengine/bevy#16093](https://github.com/bevyengine/bevy/pull/16093). As a workaround we can always just enable the `custom_cursor` feature and probably no one will notice or care, but still, it's relevant.

I implemented your suggestion by always including custom_cursor, this is closest to previous behaviour, a future PR could easily build on that to add custom cursor support for bevy_egui, but I'll consider this out of scope for this "upgrade bevy" PR.

vladbat00:
@Vrixyz huge thanks, as always! I've just updated it to the actual 0.15 release and merged.

PR #313, opened 2024-10-03
egui 0.29 support

PPakalns:
Egui added "multipass" rendering for cases when multiple passes are needed to layout UI correctly.

For now this feature can be used manually (See simple multipass example) or old behaviour can still be used by default.

In future probably begin_frame_pass, end_frame_pass systems and all user systems could be rerun. But investigations how to do set of system rerun in bevy must be done.


mvlabat:
Thank you for the PR!

I'll push a little change soon to fix the test CI pass. I'll also move `manual_run` to `EguiSettings` and convert it to a component to allow users control running per context.

PR #319, opened 2024-11-08
Compile time improvement: depend on subcrates

aevyrie:
`bevy_egui` was causing compilation of our large projects to delay until `bevy`, and more critically, `bevy_pbr` were done compiling. `bevy_pbr` is extremely slow, and depends on `bevy_render`, which is also slow to compile. The result is compilation cannot parallelize well.

This change should be entirely idempotent for users, it simply switches dependencies to the bevy subcrates. The effect of this is that bevy_egui, and anything that depend on it, can now start compiling at the same time as `bevy_pbr`, this had a huge positive effect on compile times in our larger project, because many things depend on `bevy_egui`.

The only "gotcha" here is that I had to depend on `encase` manually, otherwise you end up needing to go through bevy's `ShaderType` derive, which falls over when `bevy` is not in scope.

Using `--timings` on the `simple` example, on an M3 Max:

Before
image

After
image

As expected, this has no impact on the compile time of examples, because `bevy_pbr` still dominates, but in real applications, you will usually have a bunch of crates depending on `bevy_egui`, which can now start compiling a whole 4 seconds earlier (and even more on slower machines).

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -179,7 +179,7 @@ jobs:
           restore-keys: |
             cache-doc-cargo-${{ hashFiles('**/Cargo.toml') }}
             cache-doc-cargo
-      - run: cargo doc --all --features "bevy/x11"
+      - run: cargo doc --all --features "bevy_winit/x11"
         env:
           RUSTDOCFLAGS: -D warnings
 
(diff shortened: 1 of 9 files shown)


mvlabat:
This is great, thank you!

PR #321, opened 2024-11-24
Don't add EguiContext to every entity when render feature is disabled.

DGriffin91:
When Initial world space UI support (https://github.com/vladbat00/bevy_egui/pull/304) was added, the fallback query for when the render feature is disabled results in `EguiContext`s being added every entity in the scene.

This PR makes it so that `setup_render_to_texture_handles_system` only runs when the render feature is enabled.


vladbat00:
Oooof, I'm a bit embarrassed to not have caught that :D Thank you for the fix!

PR #323, opened 2024-11-30
Update to bevy 15.0

simon-an:
Created  based on the other pullrequest https://github.com/vladbat00/bevy_egui/pull/309

Tested with all the examples. The UI in my game looks like before moving from 14 to 15.


vladbat00:
Seems like our commits were just 5 minutes apart 😄 Already merged #309. Thank you for your work though!

PR #332, opened 2024-12-08
Fix multiple windows when a window is spawned later

jabuwu:
If multiple windows are spawned at `Startup`, they all work correctly. If a window is spawned later in `Update`, they do not work. Input works, but rendering does not.

I'm not sure if this is the right solution to this problem, as I don't understand how the new `RenderEntity` works. At the time `Window` is added, `RenderEntity` does not yet exist on the entity. However, at the time `EguiContext` is added, it does exist.


OwlyCode:
I'm encountering the issue as well, I use the F1 key to toggle a debug window in my game and it's empty after updating to bevy 0.15.

jabuwu:
It's also possible to workaround this issue by spawning windows in `PreUpdate` with `EguiContext::default()` already attached.

vladbat00:
Hi, thank you for the PR!

I think this fixes attaching egui contexts when spawning a window not on startup, but it also will create `EguiNode` node twice for image render targets (see `setup_new_render_to_image_nodes_system`) - not sure if that breaks anything but it's not ideal anyway.

That's a nice catch about `RenderEntity `. I'm not sure why it doesn't exist yet... It can only mean that the window entity doesn't receive the `SyncToRenderWorld` component in time. I'm still trying to understand when exactly it gets added. It might be as well something that should be fixed on the Bevy end (possibly some system ordering issues), but I don't know yet.

We can definitely workaround it somehow on the `bevy_egui` end but I'd like to have a deeper look into Bevy itself to understand how exactly that happens.

vladbat00:
Ok, it's definitely something we should fix on our end. The reason `SyncToRenderWorld` is missing is that it isn't supposed to be added for `Window` entities automatically. It gets added only because of the `ExtractComponentPlugin::<EguiContext>` plugin.
So if `setup_new_windows_system` runs during the `PreUpdate` schedule and creating a window happens after it, there will be 1 frame when an entity has a `Window` component but doesn't have a `EguiContext` component.

So another way for users to workaround that is to spawn a window during `PreUpdate` with `.before(setup_new_windows_system)`.

Your fix actually looks good, I'll push a little change soon to make it work correctly with image render targets as well.

jabuwu:
Thank you 🙏

Issue #283, opened 2024-05-28
Create new texture in egui makes the framerate very low


crane-may:
In my application, I need to generate a new texture each time in egui update. I found that `update_egui_textures_system` sometimes takes more than 0.5 seconds to run. Then I located the code that was causing the performance:

```rust
pub(crate) fn color_image_as_bevy_image(
    egui_image: &egui::ColorImage,
    sampler_descriptor: ImageSampler,
) -> Image {
    let pixels = egui_image
        .pixels
        .iter()
        // We unmultiply Egui textures to premultiply them later in the fragment shader.
        // As user textures loaded as Bevy assets are not premultiplied (and there seems to be no
        // convenient way to convert them to premultiplied ones), we do the this with Egui ones.
        .flat_map(|color| color.to_srgba_unmultiplied())
        .collect();

...
```
My guess is `to_srgba_unmultiplied` is causing the slowdown. Then I replaced it with the crate `fast-srgb8`, which is x10 times faster!

```rust
pub(crate) fn color_image_as_bevy_image(
    egui_image: &egui::ColorImage,
    sampler_descriptor: ImageSampler,
) -> Image {
    let pixels = egui_image
        .pixels
        .iter()
        // We unmultiply Egui textures to premultiply them later in the fragment shader.
        // As user textures loaded as Bevy assets are not premultiplied (and there seems to be no
        // convenient way to convert them to premultiplied ones), we do the this with Egui ones.
        .flat_map(|color| fast_srgb8::f32x4_to_srgb8(color.to_normalized_gamma_f32()))
        .collect();

...
```

I hope it helps if you are having this same problem.

gogo2077:
Thank you, this is very helpful.

mvlabat:
Hi! Thank you for sharing. I haven't got a chance to test it myself yet. Btw, have you tried running your app in the release mode, or at least [compiling dependencies with optimisations](https://doc.rust-lang.org/cargo/reference/profiles.html#overrides)?

Will the slowdown still be this significant?

crane-may:
> Hi! Thank you for sharing. I haven't got a chance to test it myself yet. Btw, have you tried running your app in the release mode, or at least [compiling dependencies with optimisations](https://doc.rust-lang.org/cargo/reference/profiles.html#overrides)?
> 
> Will the slowdown still be this significant?

Sorry for not paying attention to this message and replying a bit late. 

Yes, I was testing in release mode.

Aronry:
[...]

PR #334, opened 2024-12-12
Enhance performance of color_image_as_bevy_image

gogo2077:
#283 
Before

After

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -73,6 +73,9 @@ bevy_utils = "0.15.0"
 bevy_winit = { version = "0.15.0", features = ["custom_cursor"] }
 bevy_window = "0.15.0"
 
+# Enhance performance of color_image_as_bevy_image
+fast-srgb8 = "1.0.0"
+
 # `open_url` feature
 webbrowser = { version = "1.0.1", optional = true }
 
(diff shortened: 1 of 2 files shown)


vladbat00:
Hi, thanks for the PR! This is a good improvement, but I think like I'd rather have this optimisation upstreamed to `ecolor` so that we can keep using their implementation and avoid adding other dependencies (even though quite lightweight one). How do you feel about contributing this directly to Egui's `ecolor` instead?

Btw, even 33ms looks like a lot.. What's the size of the image you are trying to update? This made me think whether offloading the conversion to a separate task pool would make sense. (Not sure it would make any difference in WASM though...)

Or we can look into implementing a partial image update if that can help in your scenario.

gogo2077:
> Hi, thanks for the PR! This is a good improvement, but I think I'd rather have this optimisation upstreamed to `ecolor` so that we can keep using their implementation and avoid adding other dependencies (even though quite a lightweight one). How do you feel about contributing this directly to Egui's `ecolor` instead?
> 
> Btw, even 33ms looks like a lot.. What's the size of the image you are trying to update? This made me think whether offloading the conversion to a separate task pool would make sense. (Not sure it would make any difference in WASM though...)
> 
> Or we can look into implementing a partial image update if that can help in your scenario.



> Hi, thanks for the PR! This is a good improvement, but I think I'd rather have this optimisation upstreamed to `ecolor` so that we can keep using their implementation and avoid adding other dependencies (even though quite a lightweight one). How do you feel about contributing this directly to Egui's `ecolor` instead?
> 
> Btw, even 33ms looks like a lot.. What's the size of the image you are trying to update? This made me think whether offloading the conversion to a separate task pool would make sense. (Not sure it would make any difference in WASM though...)
> 
> Or we can look into implementing a partial image update if that can help in your scenario.

Hello, I’m not very familiar with egui, but I can give it a try.

I’m developing a card game where individual images aren’t very large, but there might be a lot of images and fonts created all at once.

PR #339, opened 2024-12-22
Update links to examples in README.md

theKlisha:
Just noticed they were out of date after navigating to one of them via readme link, and finding out that it was not compiling.


vladbat00:
Thanks! I've got a bad habit of forgetting to update these links...

PR #353, opened 2025-01-23
Fix reactive rendering regression

### LINKED ISSUE #352: Regression: clicks are missed in reactive rendering mode
When using bevy with reactive winit rendering to reduce power draw, bevy_egui will ignore button presses that wake the event loop.

This was fixed in https://github.com/mvlabat/bevy_egui/pull/226, but that change seems to have been crushed or reverted.

Bug repro:
1. Set bevy's win plugin to use low power mode
2. Launch an egui example with a button
3. Move mouse over the button
4. Wait for a few seconds
5. Click the button, nothing will happen

aevyrie:
Fixes #352

### DIFF
--- a/src/systems.rs
+++ b/src/systems.rs
@@ -359,7 +359,7 @@ pub fn process_input_system(
     }
 
     for mut context in context_params.contexts.iter_mut() {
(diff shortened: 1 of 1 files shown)


aevyrie:
image
 wat

PR #359, opened 2025-02-08
Bevy egui performance improvements

PPakalns:
1. Move wgpu::Buffer preperation to bevy render world Prepare system set. This allows parallel execution with other Prepare system set systems. If it is done in render graph, execution is serial.
2. Parallelize render command encoding.
3. Optimize extraction of EguiRenderOutput to render world.  **Huge performance gain** (More than 1ms in my case per frame).  Instead of cloning whole array, Vec is wrapped inside an Arc and just simple Clone operation is done.  This operation allows Main world and render world to resume processing instead of waiting until all data is cloned.
4. Optimize RenderPass scissor rect handling.  Bevy TrackedRenderPass doesn't track scissor rect changes and manual handling is needed to remove unnecessary commands.
5. Remove unnecessary memory allocations (Frequent small vector creation) during vertex/index buffer preparation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -129,7 +129,7 @@ use crate::text_agent::{
 #[cfg(feature = "render")]
 use crate::{
     egui_node::{EguiPipeline, EGUI_SHADER_HANDLE},
-    render_systems::{EguiTransforms, ExtractedEguiManagedTextures},
+    render_systems::{EguiRenderData, EguiTransforms, ExtractedEguiManagedTextures},
 };
 #[cfg(all(
     feature = "manage_clipboard",
@@ -1012,6 +1012,7 @@ impl Plugin for EguiPlugin {
                 .init_resource::<egui_node::EguiPipeline>()
                 .init_resource::<SpecializedRenderPipelines<EguiPipeline>>()
                 .init_resource::<EguiTransforms>()
+                .init_resource::<EguiRenderData>()
                 .add_systems(
                     // Seems to be just the set to add/remove nodes, as it'll run before
                     // `RenderSet::ExtractCommands` where render nodes get updated.
@@ -1026,6 +1027,10 @@ impl Plugin for EguiPlugin {
                 .add_systems(
                     Render,
                     render_systems::prepare_egui_transforms_system.in_set(RenderSet::Prepare),
+                )
+                .add_systems(
+                    Render,
+                    render_systems::prepare_egui_render_target_data.in_set(RenderSet::Prepare),
                 )
                 .add_systems(
                     Render,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -172,6 +172,7 @@ use output::process_output_system;
     not(any(target_arch = "wasm32", target_os = "android"))
 ))]
 use std::cell::{RefCell, RefMut};
+use std::sync::Arc;
 #[cfg(target_arch = "wasm32")]
 use wasm_bindgen::prelude::*;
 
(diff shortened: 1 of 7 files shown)


PPakalns:
This should be merged after egui 0.31 support is merged.

vladbat00:
Hi! This is awesome, thank you a lot for the PR!
I might need some time before I can review and test it, as I'm a bit too busy with IRL stuff atm, but I hope to get to the review this weekend.

src/render_systems.rs
```diff
@@ -334,3 +339,229 @@ pub fn queue_pipelines_system(
...
+    }
+}
+
+fn round_up_to_pow2(size: usize) -> usize {
+    if size.is_power_of_two() {
+        size
+    } else {
+        size.next_power_of_two()
+    }
+}
```
vladbat00, line 567: This function seems to be redundant: https://doc.rust-lang.org/std/primitive.usize.html#method.next_power_of_two
```
assert_eq!(2usize.next_power_of_two(), Some(2));
```

src/lib.rs
```diff
@@ -334,11 +335,16 @@ pub struct EguiClipboard {
 pub struct EguiRenderOutput {
     /// Pairs of rectangles and paint commands.
     ///
-    /// The field gets populated during the [`EguiPostUpdateSet::ProcessOutput`] system (belonging to bevy's [`PostUpdate`]) and reset during [`egui_node::EguiNode`]'s `update`.
-    pub paint_jobs: Vec<egui::ClippedPrimitive>,
+    /// The field gets populated during the [`EguiPostUpdateSet::ProcessOutput`] system (belonging to bevy's [`PostUpdate`])
+    /// and processed during [`egui_node::EguiNode`]'s `update`.
+    ///
+    /// Value is wrapped in [`Arc`] to improve [`ExtractComponent`] performance.
+    pub paint_jobs: Arc<Vec<egui::ClippedPrimitive>>,
```
nsabovic, line 342: I don't know what's the etiquette for drive-by commenting on random projects on the internet but:

`Arc<[egui::ClippedPrimitive]>` would avoid double allocation and I don't see the array being appended to after being created.

Apologies if this is unwanted nitpicking.

PPakalns: This introduces additional full array memory copy. 
Because slice in Arc<[T]> can contain only initialized members and Vec may contain additional capacity. 
See source code: https://doc.rust-lang.org/src/alloc/sync.rs.html#3719
Maybe some compiler optimizations can optimize this (but I am not sure because memory allocators are involved), but current approach would work even in debug builds.

Moving Vec<_> into Arc<Vec<_>> guarantees that underlaying Vec memory is not touched, but only Vec structure that contains pointer, capacity, used length is copied into Arc.  This is a lot faster.

PPakalns: Checked in compiler explorer.

In suggested appraoch more allocation / memcpy operations are introduces:
https://godbolt.org/z/zx7js9E4q

vladbat00: I guess it would be possible to avoid copies/reallocations if Egui itself returned paint jobs as `Box<[egui::ClippedPrimitive]>` (but I'm not sure if there's a way to cast `Box` to `Arc` without copying) or just `Arc<[egui::ClippedPrimitive]>`. And I imagine that would involve some unsafe machinery to initialise such an array on the Egui end as well, which would probably bring only negligible gains anyway.

`Arc<Vec<egui::ClippedPrimitive>>` seems like the best option for now.

vladbat00:
I've just tested it and confirmed the performance gains, thank you! When I ran the `ui` example on my mac, extracting render output was about 16 microseconds, with the changes in your PR it's now like a couple of microseconds. Probably not the best example to test the performance difference, but it is there! 😄 

Thank you again!

Issue #218, opened 2023-10-10
Add run condition for whether egui is using pointer input


jakobhellermann:
It came up in https://github.com/jakobhellermann/bevy-inspector-egui/pull/147 that it is often useful to run some systems (like looking around) only if egui is not using the pointer input.

 

```rs
mod common_conditions {
  fn egui_mouse_unused(egui_contexts: Query<&EguiContext>) -> bool {
      egui_contexts
          .iter()
          .all(|ctx| !ctx.get().wants_pointer_input())
  }
}
```

```rs
app.add_systems(
    Update,
    (camera_pan, camera_zoom).run_if(bevy_egui::common_conditions::mouse_pointer_valid()),
)
```

For prior art, bevy also has some `common_conditions` for various subcrates:
- https://docs.rs/bevy/latest/bevy/time/common_conditions/index.html
- https://docs.rs/bevy/latest/bevy/input/common_conditions/index.html
- https://docs.rs/bevy/latest/bevy/ecs/schedule/common_conditions/index.html

PR #369, opened 2025-03-26
Implement absorbing inputs and Egui input run condition

### LINKED ISSUE #47: Absorbing/suppressing keyboard/mouse input when hovering over egui window

vladbat00:
Closes #47, #218

- Add `egui_wants_input` run condition
- Add `absorb_bevy_input_system` system (disabled by default, can be enabled with `EguiGlobalSettings::enable_absorb_bevy_input_system`)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,9 @@ serde = ["egui/serde"]
 log_input_events = []
 
 [[example]]
+name = "absorb_input"
+required-features = ["render"]
+[[example]]
 name = "color_test"
 required-features = ["render"]
 [[example]]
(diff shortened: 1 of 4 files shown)


philpax (approved):
Looks pretty reasonable, great work! Maybe add some comments as to why it isn't enabled by default?

PR #375, opened 2025-04-08
Fix missing egui file drag and drop events

MickHarrigan:
## Description
 
This PR adds functionality to pass `bevy_window::FileDragAndDrop` events down to egui. Additionally adds a bevy-ified version of the [egui file_dialog example](https://github.com/emilk/egui/blob/master/examples/file_dialog/src/main.rs). 

 ## Changes:
 
* Added a `log_file_dnd_events` feature similar in functionality to the `log_input_events` feature
* Added an example of bevy events going through to egui for files being dragged and dropped onto the window
* Added new bevy systems, event, and setting for controlling event passthrough

## Why this is needed:

Currently when using `bevy_egui` any files that get dragged and dropped (or hovered) onto the window can only be read from directly via bevy. With this it allows passing down into the egui context for usage there.

## Limitations:

I don't believe that there is a concise way to handle the location of a file drop / hover, only that it exists.

This is to be fixed with the release of `winit 0.31` afaik and should eventually be covered.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,9 +37,6 @@ log_input_events = []
 # The enabled logs will print with the info log level, to make it less cumbersome to debug in browsers.
 log_file_dnd_events = []
 
-# specifically for feature gating `rfd` from tests
-file_browse = ["rfd"]
-
 [[example]]
 name = "absorb_input"
 required-features = ["render"]
@@ -75,7 +72,7 @@ name = "render_egui_to_image"
 required-features = ["picking", "render", "bevy/bevy_gizmos"]
 [[example]]
 name = "file_browse"
-required-features = ["render", "file_browse"]
+required-features = ["render"]
 
 [dependencies]
 egui = { version = "0.31", default-features = false }
@@ -110,10 +107,6 @@ bevy_picking = { version = "0.16.0-rc", optional = true }
 [target.'cfg(not(any(target_arch = "wasm32", target_os = "android")))'.dependencies]
 arboard = { version = "3.2.0", optional = true }
 thread_local = { version = "1.1.0", optional = true }
-
-# `file_browse` feature only, workaround dev-dependency
-[target.'cfg(not(any(target_os = "android", target_os = "ios", target_arch = "wasm32")))'.dependencies]
-rfd = { version = "0.13", optional = true }
 
 [dev-dependencies]
 version-sync = "0.9.5"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,9 @@ log_input_events = []
 # The enabled logs will print with the info log level, to make it less cumbersome to debug in browsers.
 log_file_dnd_events = []
 
+# specifically for feature gating `rfd` from tests
+file_browse = ["rfd"]
+
 [[example]]
 name = "absorb_input"
 required-features = ["render"]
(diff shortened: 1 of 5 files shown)


src/input.rs
```diff
@@ -811,6 +877,62 @@ pub fn write_egui_input_system(
...
+        }
+    }
+
+    for (entity, mut egui_input, window) in egui_contexts.iter_mut() {
+        egui_input.focused = focused_non_window_egui_context.as_deref().map_or_else(
+            || window.is_some_and(|window| window.focused),
+            |context| context.0 == entity,
+        );
+        egui_input.time = Some(time.elapsed_secs_f64());
+    }
```
vladbat00, line 933: I don't think we should do this in both `write_egui_input_system` and `write_egui_file_dnd_system`. In fact, I think we can just merge the dnd-relevant code into `write_egui_input_system`.

vladbat00 (commented):
Hi, thank you for the PR, this is a very welcome addition! I added a minor comment, but apart from that and failing CI for some reason, the PR looks good to me.

src/input.rs
MickHarrigan: Addressed with commit [53df2e1](https://github.com/vladbat00/bevy_egui/pull/375/commits/53df2e1a143a02ab2610b48230f17ed88f5c827e)

MickHarrigan:
> Hi, thank you for the PR, this is a very welcome addition! I added a minor comment, but apart from that and failing CI for some reason, the PR looks good to me.

With regards to the CI stuff, I have a commit ready that would put the relevant code behind a `cfg(...)` that would maintain the current functionality on Windows and Unix but do nothing whatsoever otherwise.

Is that a preferable action for this? I am not well versed with wasm, ios, nor android (and I think the latter 2 wouldn't support file selection easily?) so I am curious to hear your opinion.

vladbat00:
Yes, let's just disable the code for those targets.

MickHarrigan:
Got CI checks just about working but running `cargo test --all` causes an error for running out of disk space. I am not great with CI myself and have been looking for a way around the disk usage but haven't gotten anything yet :sweat_smile:

vladbat00:
Not sure what's the best way to deal with it, but I usually go to the cache page and just nuke everything one by one :D
In your case that would be https://github.com/MickHarrigan/bevy_egui/actions/caches

Cargo.toml
```diff
@@ -103,6 +111,10 @@ bevy_picking = { version = "0.16.0-rc", optional = true }
 arboard = { version = "3.2.0", optional = true }
 thread_local = { version = "1.1.0", optional = true }
 
+# `file_browse` feature only, workaround dev-dependency
+[target.'cfg(not(any(target_os = "android", target_os = "ios", target_arch = "wasm32")))'.dependencies]
+rfd = { version = "0.13", optional = true }
```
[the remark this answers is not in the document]
vladbat00: Oh, I can see this was even your initial version. I assume you were trying to workaround the CI weirdness?..

So basically [here](https://github.com/MickHarrigan/bevy_egui/commit/be20862d1f12bcedddf649d1f3aa7a1769db6c36) the build was already failing due to the out of space error: https://github.com/MickHarrigan/bevy_egui/actions/runs/14431554045/job/40467118849

So maybe we can partially revert [this one](https://github.com/vladbat00/bevy_egui/pull/375/commits/eb84bb54c311c268a6a65e293800919a6fcdb103)?

vladbat00:
I've merged your PR via #377 - it takes your branch except for the last two commits. That's basically the code I wanted to see, sorry if I my comments were confusing 😄 
Thanks a lot for your work on this!

PR #380, opened 2025-04-27
Remove default features from `image` dependency

### LINKED ISSUE #379: `bevy_egui` pulls `image` with default `avif` feature, requires non-Bevy-approved license for `libfuzzer-sys`
`bevy_egui` pulls `image` [without `default-features = false`](https://github.com/vladbat00/bevy_egui/blob/47e26f141ed1a78d266586a49a8e6cb59436a954/Cargo.toml#L98) (unlike [Bevy's own `bevy_image`](https://github.com/bevyengine/bevy/blob/7f0490655c3ede945e6babc1f1a1f06b13404aa8/crates/bevy_image/Cargo.toml#L57)). This activates the default `avif` feature of `image`, which in turn pulls the `ravif` dependency which ends up pulling `libfuzzer-sys`. That library has an SPDX license of "(MIT OR Apache-2.0) AND NCSA". The `NCSA` part is not [a Bevy approved license](https://github.com/bevyengine/bevy/blob/7f0490655c3ede945e6babc1f1a1f06b13404aa8/deny.toml#L15-L28).

I appreciate there's no requirement for `bevy_egui` to follow the exact same license rules as Bevy itself, but I'm guessing this is more of an oversight than a conscious decision maybe? And this currently prevents me from upgrading Hanabi, which pulls `bevy_egui` and uses the same license rules as Bevy (Hanabi's CI fails on `cargo deny` rule). I can work around, but opening this issue to ask if you'd consider fixing this and aligning with Bevy's own license rules to make things more consistent?

djeedai:
Import the `image` dependency without its default features, for consistency with `bevy_image` and to avoid the `avif` feature which pulls some NCSA license, which is currently not one of the licenses Bevy itself approves. This makes it more consistent with Bevy.

Fixes #379

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -149,7 +149,7 @@ web-sys = { version = "0.3.74", features = [
     "TouchEvent",
     "Window",
 ] }
-image = { version = "0.25.5", features = ["png"] } # For copying images
+image = { version = "0.25.5", default-features = false, features = ["png"] } # For copying images
 js-sys = "0.3.63"
 wasm-bindgen = "0.2.93"
 wasm-bindgen-futures = "0.4.36"


vladbat00:
Thank you!

PR #383, opened 2025-05-17
Show the virtual keyboard on mobile when text input is focused

arcln:


arcln:
So after using this PR in my project, I noticed it creates a werid bug where the touch inputs completely stops working. I am investigating this issue, do not merge this PR until I fix this.

vladbat00:
Thanks for the PR!

Issue #358, opened 2025-02-05
Handling request_discard?


OwlyCode:
Hi!

I'm using egui's Grids into my UI, and I've noticed that each time a new grid is rendered, there is a noticeable "glitchy" frame occuring. I've tracked this down to the fact that adding a new grid trigger a request discard:

```rust
// grid.rs:437

// The initial frame will be glitchy, because we don't know the sizes of things to come.

if ui.is_visible() {
    // Try to cover up the glitchy initial frame:
    ui.ctx().request_discard("new Grid");
}
```

It seems that when this is called, all rendering will stop (because the egui context switches to a special mode), causing the "glitchy" frame. I'm not sure on how this could be handled because all rendering happens in systems and we can't just re-run them. Would it be possible in that case to re-use the rendering from the previous frame?

Here's how it looks:

![Image](https://github.com/user-attachments/assets/8d181d16-9a8c-4cdb-906f-1aa9c2f61cc5)

![Image](https://github.com/user-attachments/assets/53a17c1c-b3d3-4b10-aad3-a4fc024b7241)

vladbat00:
Hi! Sorry for the delayed response. You could try calling `egui::Context::run` manually and put all of your systems' code there. It handles discard by looping the passed `run_ui` lambda until the discards stop or the `egui::Option::max_passes` limit is reached. To use `bevy_egui` like that, you'll also need to set `EguiContextSettings::run_manually` to `true`.

See https://github.com/vladbat00/bevy_egui/blob/d351599b9a266eb52733eb842f535289dffd6297/examples/simple_multipass.rs.

OwlyCode:
Hi! No problem I'm already lucky to have this library!

I'll see if I can do that. For now it was easier for me to skip using grids entirely: I'd have around 40 UI systems to move to achieve that.

vladbat00:
I see. You might also find `World::run_system` useful, that way you'd have a `&mut World` system that would run your existing UI systems inside, without having to move too much code around. That basically involves only moving your UI systems from bevy app/schedule setup to the `&mut World` system.

vladbat00:
Hi! It seems there's a good chance we'll have `request_discard` support in the next release :)

The PR is still work-in-progress (mainly documentation-wise, the functionality is already there): #370. But I'll appreciate if you can try it out with your example. And if you have any feedback/review comments, feel free to leave them in the PR ofc!

OwlyCode:
[...]

Issue #212, opened 2023-09-28
`EguiContexts::ctx_mut` occasionally causes a panic when closing the app.


joseph-gio:
```
thread 'Compute Task Pool (2)' panicked at '`EguiContexts::ctx_mut` was called for an uninitialized context (primary window), make sure your system is run after [`EguiSet::InitContexts`] (or [`EguiStartupSet::InitContexts`] for startup systems)', /home/joseph/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bevy_egui-0.21.0/src/lib.rs:337:14
```

This panic occurs intermittently when closing my app. Based on the panic message, I'm assuming that the `EguiContext` for the primary window is being deinitialized before the app fully closes, after which my system is allowed to run, which fails since it's trying to access an uninitialized context.

Backtrace: 

```
stack backtrace:
   0: rust_begin_unwind
             at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/core/src/panicking.rs:67:14
   2: core::panicking::panic_display
             at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/core/src/panicking.rs:150:5
   3: core::panicking::panic_str
             at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/core/src/panicking.rs:134:5
   4: core::option::expect_failed
             at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/core/src/option.rs:1952:5
   5: core::option::Option<T>::expect
             at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/core/src/option.rs:898:21
   6: bevy_egui::EguiContexts::ctx_mut
             at /home/joseph/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bevy_egui-0.21.0/src/lib.rs:333:47
   7: creator::home::show_loading_screen
             at ./src/home.rs:38:15
   8: core::ops::function::FnMut::call_mut
             at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/core/src/ops/function.rs:166:5
   9: core::ops::function::impls::<impl core::ops::function::FnMut<A> for &mut F>::call_mut
[...]

PR #387, opened 2025-05-27
Deprecate the option to disable the multi-pass mode

vladbat00:
This PR deprecates the option to disable the multi-pass mode. Ideally, that's the only mode I'd like to support in `bevy_egui`, as it helps to keep systems organised in a single schedule (`EguiContextPass`), which in turn helps to avoid issues such as #212, makes widgets requiring multiple passes work by default (#358), etc.

Also, I hope that the deprecation will prompt users to report issues with the multi-pass mode if there are any. If there are issues with the multi-pass mode that can't be easily solved or there are any reasons why users want to keep using the single-pass mode (or, for example the `Update` schedule specifically), we'll at least learn of those. (And there's always an option to "un-deprecate" the single-pass mode in that case.)

Having said that, my intention is to keep the single-pass mode support around for Bevy 0.16-0.17, and I hope to remove it completely with Bevy 0.18.

### DIFF
--- a/examples/absorb_input.rs
+++ b/examples/absorb_input.rs
@@ -15,9 +15,7 @@ use bevy_egui::{
 fn main() {
     App::new()
         .add_plugins(DefaultPlugins)
-        .add_plugins(EguiPlugin {
-            enable_multipass_for_primary_context: true,
-        })
+        .add_plugins(EguiPlugin::default())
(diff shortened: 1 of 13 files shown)


jnhyatt:
Nice, happy to see the default impl!

PR #392, opened 2025-06-23
Attach `EguiContext` instances to cameras

vladbat00:


vladbat00:
@PixelDust22 sorry for the inconvenience and thanks for calling the migration issue out.

From my end, I tried to prepare users to it, but I understand the info is still insufficient and users do have issues during the migration. The change was mentioned in the changelog and highlighted as a breaking one, it was also noted that users need a camera to be able to render Egui, and the note referenced an example.
I understand it's not an excessive migration guide but it was not something feasible for me to do (to cover all possible cases where it would affect users).
I realise now it should also include a hint that if an app uses several cameras, users also need additional changes, such as marking the context camera manually (I can add that soon).

Several cameras - is it the same case for you? If you can share what exactly would make the migration easier for you, that would help me and users that are yet to perform it (was it difficult to discover the info about the breaking change? where would you look for the info first? what problem exactly did you have and what piece of info was missing for you?)

PR #399, opened 2025-07-12
update to egui 0.32

hacknus:
I removed
```rust
fn alpha_image_as_color_image(image: &egui::FontImage) -> egui::ColorImage {
    egui::ColorImage {
        size: image.size,
        pixels: image.srgba_pixels(None).collect(),
    }
}
```
sind `ImageData::Font(image)` has been removed.

but I still have the issue of `wants_repaint_after()` which is now private:
```rust
error[E0624]: method `wants_repaint_after` is private
   --> src/output.rs:120:56
    |
120 |                 ctx.viewport(|viewport| viewport.input.wants_repaint_after())
    |                                                        ^^^^^^^^^^^^^^^^^^^ private method
    |
   ::: /Users/linus/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/egui-0.32.0/src/input_state/mod.rs:607:5
    |
607 |     pub(crate) fn wants_repaint_after(&self) -> Option<Duration> {
    |     ------------------------------------------------------------ private method defined here
```

any ideas?

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -116,11 +116,11 @@ pub fn process_output_system(
             // winit, that it needs to wake up next frame as well even if there are no inputs.
             //
             // TLDR: this solves repaint corner cases of `WinitSettings::desktop_app()`.
-            if let Some(Duration::ZERO) =
-                ctx.viewport(|viewport| viewport.input.wants_repaint_after())
-            {
-                let _ = event_loop_proxy.send_event(WakeUp);
-            }
+            // if let Some(Duration::ZERO) =
+            //     ctx.viewport(|viewport| viewport.input.wants_repaint_after())
+            // {
+            //     let _ = event_loop_proxy.send_event(WakeUp);
+            // }
         }
     }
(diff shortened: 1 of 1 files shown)


hacknus:
Can someone confirm if this is still required?

```rust
        if let Some(event_loop_proxy) = &event_loop_proxy {
            // A zero duration indicates that it's an outstanding redraw request, which gives Egui an
            // opportunity to settle the effects of interactions with widgets. Such repaint requests
            // are processed not immediately but on a next frame. In this case, we need to indicate to
             // winit, that it needs to wake up next frame as well even if there are no inputs.
             //
             // TLDR: this solves repaint corner cases of `WinitSettings::desktop_app()`.
             if let Some(Duration::ZERO) =
                 ctx.viewport(|viewport| viewport.input.wants_repaint_after())
             {
                 let _ = event_loop_proxy.send_event(WakeUp);
             }
         }
     
```

PR #402, opened 2025-07-14
Update egui to 0.32

globin:
This removes a WakeUp event that was sent if wants_repaint_after returned a zero duration. The function is now only pub(crate) and can no longer be used, possibly this needs a new workaround.


globin:
I just noticed that I missed #399, this essentially does the same and additionally fixes up some deprecation warnings in examples.

PR #405, opened 2025-07-22
Configure render ordering vis-a-vis Bevy UI

### LINKED ISSUE #404: Rendering order with bevy-ui is inconsistent
I would expect bevy_egui and bevy::ui to be able to coexist in the same project, and have a consistent z-ordering between them (_probably_ with bevy_egui in front, assuming that projects using both would use egui for debugging overlays, and bevy::ui for the "primary" game UI).

Instead, it seems to be a 50/50 coin flip on each launch about which one will end up in front.

For example, try `cargo run`ning https://codeberg.org/natkr/repros/src/branch/bevy-egui/unpredictable-ui-order a few times... you should see both Bevy "winning" the race:

Image

And egui:

Image

nightkr:
Fixes #404

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -367,6 +367,7 @@ impl Default for EguiPlugin {
         Self {
             #[allow(deprecated)]
             enable_multipass_for_primary_context: true,
+            #[cfg(feature = "bevy_ui")]
             ui_render_order: UiRenderOrder::EguiAboveBevyUi,
         }
     }
@@ -376,6 +377,7 @@ impl Default for EguiPlugin {
 ///
 /// See [`EguiPlugin::ui_render_order`].
 #[cfg(feature = "bevy_ui")]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum UiRenderOrder {
     /// [`egui`] UIs are rendered on top of [`bevy_ui`](Bevy UI).
     EguiAboveBevyUi,


src/lib.rs
```diff
@@ -1283,6 +1302,50 @@ impl Plugin for EguiPlugin {
...
+                if let Some(graph_2d) =
+                    graph.get_sub_graph_mut(bevy_core_pipeline::core_2d::graph::Core2d)
+                {
+                    // Only apply if the bevy_ui plugin is actually enabled.
+                    // In theory we could use RenderGraph::try_add_node_edge instead and ignore the result,
+                    // but that still seems to end up writing the corrupt edge into the graph,
+                    // causing the game to panic down the line.
+                    if graph_2d
+                        .get_node_state(bevy_ui::graph::NodeUi::UiPass)
+                        .is_ok()
```
vladbat00, line 1333: I think we should add a warning here if we fail to add the edge, to explain to a user why their config hasn't been applied

vladbat00:
Great work, thank you! Added a small suggestion, looks perfect to me otherwise

src/lib.rs
nightkr: Yeah I was going back and forth on this.. it's a weird and kind of surprising error condition.. but on the other hand, that would end up logging warnings for every project that disables bevy_ui (which seems like a natural thing to do if you use egui for all your UI needs).

Cargo.toml
```diff
@@ -15,7 +15,7 @@ features = ["bevy_winit/x11", "immutable_ctx"]
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 [features]
-default = ["manage_clipboard", "open_url", "default_fonts", "render", "picking"]
```
nightkr, line 18: One minor gotcha here is that `bevy_ui` depends on `render`.. We might want to either split it into a separate feature or move to using explicit optional-dependency features (`dep:`).

Then again, this is arguably pretty benign.. if you're not rendering then there is no order conflict anyway.

vladbat00:
Merged, thanks again!

PR #416, opened 2025-09-09
Bindless texture support

PPakalns:
Basic support for: https://github.com/vladbat00/bevy_egui/issues/368

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -218,7 +218,6 @@ use output::process_output_system;
     not(any(target_arch = "wasm32", target_os = "android"))
 ))]
 use std::cell::{RefCell, RefMut};
-use std::num::NonZero;
 #[cfg(target_arch = "wasm32")]
 use wasm_bindgen::prelude::*;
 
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -142,21 +156,23 @@ rfd = "0.15"
...
+  "CompositionEvent",
+  "DataTransfer",
+  "Document",
+  "EventTarget",
+  "HtmlInputElement",
+  "InputEvent",
+  "KeyboardEvent",
+  "Navigator",
+  "TouchEvent",
+  "Window",
```
vladbat00, line 171: People with different toml formatter settings keep diff-fighting over this one :D
I'll be happier if the changes don't include this :)

vladbat00:
Hi, thanks so much! I'll give it a better review next week, hopefully, we can merge it before the next Bevy release

Cargo.toml
PPakalns: Formatting changes reverted

vladbat00:
Looks good to me, thank you! I wasn't able to verify performance gains, but that's probably due to the fact that the existing examples just don't have a lot of custom textures. At least, there are no regressions :)

Merged ❤️

PR #410, opened 2025-08-08
chore: bevy 0.17.0

fundon:
Upgrade to bevy 0.17.0

PR #422, opened 2025-10-01
Bevy 0.17

vladbat00:
Based on #410

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,3 @@
+# These are supported funding model platforms
+
+patreon: vladbat00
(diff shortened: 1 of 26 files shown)


PPakalns:
For texture cleanup what is the intended way now when Weak handles doesn't exist anymore.

It looks like we must clean up them ourselves through remove_image?

vladbat00:
Oh, haven't noticed that, thanks! But I think we don't need any changes in our code. Bevy will clean assets up once it sees no strong handles itself, won't it? But def need to update the documentation to remove the weak handle mentions.

PPakalns:
In my case I have many Handles and I like that bevy can clean up them in my place when i drop them all.  Current bevy_egui behaviour doesn't respect that. I think for all other parts of engine they probably use AssetId under the hood for places where they expect that textures may have already been dropped.

I am little sad that egui::TextureId doesn't work as Handle :laughing: 

Maybe we need  Handle<egui::TextureId>  :laughing:

vladbat00:
Ah, I see, good point! Yeah, we should allow passing `AssetId` as an argument. I guess custom egui handles it is then :D

I guess I'll just make a new type:
```
pub enum EguiUserTexture<T> {
  Strong(Handle<T>),
  Weak(AssetId<T>),
}
```

PR #425, opened 2025-10-02
attempt to fix spurious panic when closing window

jakobhellermann:
Attempts to fix https://github.com/bevyengine/bevy/issues/21318.

I haven't validated that this works, as I was only getting the panic occasionally.
Maybe there's a better fix where bevy_egui tracks if the windows still exists better. Or the system just needs a `after/before` constraint.

But this should be good enough probably.


vladbat00:
Looks good to me, thank you!

PR #437, opened 2025-10-23
Add support for egui built-in ctrl+/- and other zoom behaviours

TeamDman:
Egui provides some built-in zoom factor manipulation.

Egui includes:
- ctrl+'+'
- ctrl+'-'
- ctrl+'0'
- `zoom_menu_buttons(ui);`

When using the `ui.rs` example before applying this change, when using the built-in egui scaling hotkeys, the UI flickers for one frame but there is no resulting scale change.

These proposed changes add support for changes to egui zoom_factor being propagated back to the bevy_egui scale_factor.

The changes include adding `ctrl+[` and `ctrl+]` hotkeys to the `ui.rs` example to demonstrate that both egui and bevy_egui can harmoniously adjust the zoom factor with no one-frame problems.

To experience the changes, run the ui.rs example and use the `ctrl+[`, `ctrl+]`, `ctrl+'+'`, `ctrl+'-'` hotkeys to observe scale changes being initiated from both egui and bevy_egui.

It's a bit fiddly to make sure that the values are synced properly while allowing either to initiate a change with no flickering lol

My understanding is that the bevy_egui scale factor was introduced before egui had its own scale factor, so there's future opportunity to remove the bevy_egui scale factor entirely, but until then this should be a good middle ground.

related: https://github.com/vladbat00/bevy_egui/issues/244

### DIFF
--- a/examples/ui.rs
+++ b/examples/ui.rs
@@ -108,8 +108,7 @@ fn update_ui_scale_factor_system(
         };
         egui_settings.scale_factor = scale_factor;
     }
-    
-    
+        
     // Handle Ctrl+] (zoom in) and Ctrl+[ (zoom out)
     // Using different keybinds than egui's default Ctrl+Plus/Minus to avoid conflicts
     let ctrl_pressed = keyboard_input.pressed(KeyCode::ControlLeft)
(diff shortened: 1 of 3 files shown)


TeamDman:
I tried removing the scale factor entirely, but it wasn't as straightforward as I hoped, will keep this PR open

Issue #460, opened 2026-01-08
No Input when using custom Event for the EventLoop


RobinKellnerVector:
Hi, I've noticed that the system for collecting all `EguiContext`s is skipped when the default `EventLoop` wake event is overridden, which causes all input to be ignored. Manually adding each new `EguiContext` to the `WindowToEguiContextMap` works as a workaround. 

`event_loop_proxy` should probably be changed into an `Option<Res>`.

https://github.com/vladbat00/bevy_egui/blob/d8aeaf428c073d157ea2746dd0fb5365d24abd39/src/input.rs#L168-L175

RobinKellnerVector:
👍I'll create a hotfix

vladbat00:
cheers!

PR #461, opened 2026-01-08
Fix no input with custom EventLoop Event (#460)

RobinKellnerVector:
Hotfix for #460


vladbat00:
Thank you for the fix! I'll merge regardless of the failing test, that linker error seems to be unrelated

PR #467, opened 2026-02-13
Support trackpad pinch-to-zoom on macOS

stargazing-dino:
Hiya, this PR adds `write_pinch_gesture_messages_system` - which reads bevy's PinchGesture messages and forwards it onto egui. The formula matches what I found [here on egui-winit](https://github.com/emilk/egui/blob/64a96ef3917d4e305a79fa5f460af5fc73eabcf8/crates/egui-winit/src/lib.rs#L488).

Tested fully on my mac machine and pinching works for me

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -11,6 +11,7 @@ use bevy_ecs::{
 };
 use bevy_input::{
     ButtonInput, ButtonState,
+    gestures::PinchGesture,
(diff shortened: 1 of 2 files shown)


vladbat00:
Hi! Apologies for not reviewing it for such a long while.
Looks great, thank you for the PR!

PR #468, opened 2026-03-06
Replace custom rendering pipeline with egui-wgpu

### LINKED ISSUE #324: Why does this crate not use `egui-wgpu` and `egui-winit`
Like the title says, why is this crate not using the official helper libraries `egui-wgpu` and `egui-winit`? I use them in my own renderer project and the integration with egui becomes tiny. I understand that bevy_egui does a bit more things like exposing a way to use `Handle<Image>` but it seems like the crate could be a lot simpler if it uses those helper crates.

This bevy PR would make the integration with egui-winit even simpler https://github.com/bevyengine/bevy/pull/15884

frewsxcv:
This pull request replaces bevy_egui's custom rendering pipeline (WGSL shader, pipeline specialization, bind groups, vertex/index buffers, bindless optimization) with the official egui-wgpu crate.

This reduces maintenance burden and stays closer to the reference egui rendering implementation.

Breaking changes:

- `EguiPlugin::bindless_mode_array_size` removed
- `EguiBevyPaintCallbackImpl::render()` takes `&mut wgpu::RenderPass<'static>` instead of `&mut TrackedRenderPass<'pass>`
- `EguiBevyPaintCallbackImpl` methods receive `TextureFormat `instead of `EguiPipelineKey`
- `EguiPipelineKey`, `EguiManagedTextures`, `EguiRenderSettings`, `EguiPipeline` removed from public API

Closes #324


vladbat00:
Hi! Thanks for the PR.

I've just tested it on every example, it seems that it works ok in the examples where only a single context is used, but it's broken in every example that uses more than 1 egui context.

<details><summary>Example</summary>
image
</details> 

You've mentioned you used Claude for this, and I'll be honest, this is another big reason why I'm hesitant to merge the PR atm. Ideally, I'd rewrite this by hand while re-using the PR as a reference.

But apart from the AI topic, there's another concern I have: using `egui-wgpu` will restrict us from making new `bevy_egui` releases in case if the integration updates its `wgpu` version while we'd be still waiting for a new Bevy release. And vice versa: if Bevy releases with a newer `wgpu` version earlier than `egui-wgpu`, then we won't be able to release `bevy_egui` (until the `egui-wgpu` bump) at all - which is even worse. Considering how many projects depend on `bevy-inspector-egui` or `bevy_egui` directly, that might hinder the adoption of newer Bevy versions quite hard, which seems to me like a big dealbreaker here. :(

frewsxcv:
> I've just tested it on every example, it seems that it works ok in the examples where only a single context is used, but it's broken in every example that uses more than 1 egui context.

Good find! My applications only use one context so I didn't encounter that. I'll look into a fix later.

> You've mentioned you used Claude for this, and I'll be honest, this is another big reason why I'm hesitant to merge the PR atm. Ideally, I'd rewrite this by hand while re-using the PR as a reference.

Totally understandable! No hard feelings if you want to close this and wait for a handwritten version.

> But apart from the AI topic, there's another concern I have: using egui-wgpu will restrict us from making new bevy_egui releases in case if the integration updates its wgpu version while we'd be still waiting for a new Bevy release. And vice versa: if Bevy releases with a newer wgpu version earlier than egui-wgpu, then we won't be able to release bevy_egui (until the egui-wgpu bump) at all - which is even worse.

Great callout. For my particular use case, I can afford to wait for both upstream dependencies to bump wgpu before doing a new release, but completely understand your hesitation.

vladbat00:
Yeah, I'm inclined to close this. I didn't even think about the possible dependency hell before looking at the PR, so apologies it didn't come up earlier in the original issue. But at least now we know that when wgpu releases a stable version, it shouldn't be too painful to migrate to egui-wgpu. So thanks for taking a look at this anyway!

PR #476, opened 2026-04-23
Cargo.toml: enable "wayland-data-control" feature for arboard

bkueng:
This enables copy-paste support under wayland.

Tested with Fedora KDE.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -128,6 +128,8 @@ bevy_picking = { version = "0.18.0", optional = true, features = ["mesh_picking"
 [target.'cfg(not(any(target_arch = "wasm32", target_os = "android")))'.dependencies]
 arboard = { version = "3.2.0", optional = true }
 thread_local = { version = "1.1.0", optional = true }
+[target.'cfg(target_os = "linux")'.dependencies]
+arboard = { version = "3.2.0", optional = true, features = ["wayland-data-control"] }
 
 [dev-dependencies]
 version-sync = "0.9.5"


vladbat00:
Hi, thanks for the PR! Unfortunately, I don't have a Linux PC so I can't test this on my own, but I'm wondering how this will affect non-wayland users

Cargo.toml
```diff
@@ -128,6 +128,8 @@ bevy_picking = { version = "0.18.0", optional = true, features = ["mesh_picking"
 [target.'cfg(not(any(target_arch = "wasm32", target_os = "android")))'.dependencies]
 arboard = { version = "3.2.0", optional = true }
 thread_local = { version = "1.1.0", optional = true }
+[target.'cfg(target_os = "linux")'.dependencies]
```
Friz64, line 131: I think this should be 

```
[target.'cfg(all(unix, not(any(target_os="macos", target_os="android", target_os="emscripten"))))'.dependencies]
```

so that it also works on non-Linux operating systems. This is what arboard does: https://github.com/1Password/arboard/blob/0a73f82d454eff18f3423c8fd9b89477a17281aa/Cargo.toml#L80

vladbat00: I'm not sure I like the fact that even users of different backends will get this dependency. Maybe we should either expose our own `arboard-wayland-data-control` feature, or just let users specify the following in their `Cargo.toml`:
```
arboard = { version = "3.2.0", features = ["wayland-data-control"] }
```

And I think I prefer the latter even more

bkueng: I'm fine with any of these solutions. I can also try to test it on a non-wayland linux.
My thinking was that wayland is pretty much the standard now, and it took me a bit to figure out why copy-paste did not work. I'd like to spare that to other users.

Friz64: > just let users specify the following in their Cargo.toml

No user of this crate is ever going to remember to do that.

I propose a "manage_clipboard_wayland" feature that is enabled by default.

vladbat00: > No user of this crate is ever going to remember to do that.

Yeah, I agree. Maybe we can document it in the README, but I see how it can still be a frustration to find the info about it.

> I propose a "manage_clipboard_wayland" feature that is enabled by default.

I also like this, but is it possible to have a default feature that activates dependencies existing only for certain targets? I haven't tried it but I have some doubts if that'll work. If it does work though (and the build doesn't break for other systems), I'm happy to stick with this approach.

Friz64: Hm, I actually believe that this is a non-issue because enabling `manage_clipboard_wayland` on non-applicable targets will just not do anything. arboard is responsible for pulling in the dependency, and it only does so on the correct target.

vladbat00: This doesn't seem to be the case:
https://github.com/vladbat00/bevy_egui/actions/runs/24850481102/job/72776942403 (ran against https://github.com/vladbat00/bevy_egui/commit/8d666f1654aa742dc9fcc78164d196d90d45a21f)

Friz64: oh, that's unfortunate. arboards cfg-gate is broken for iOS :/

we really can't fix this on our side without ugly workarounds because cargo doesn't have target-specific features...

vladbat00: Yeah, which is annoying... I'm almost fine with just adding this feature for every linux user, but leaving no way to opt out still bothers me a bit. I don't know how linux app devs usually handle it: do they support both x11 and wayland and bring in both dependencies or create separate binaries?

Or, ideally, maybe `arboard` will fix the cfg gates.

Friz64: You support both in the same binary. I've never seen anyone do separate binaries. But yes, Wayland is the standard now, so I think it's fine to enable it for everyone.

Friz64: Another option might be to copy what `egui-winit` does, which is not to use arboard's Wayland support at all, and instead rely on `smithay-clipboard`.

bkueng: @vladbat00 did you decide how you want to approach this?

Friz64: fwiw in my opinion the correct solution would be to copy `egui-winit`

vladbat00: I agree, let's adopt the `egui-winit` approach, since we do use it as a reference implementation in many other aspects

bkueng: Sounds good. I might have some time to look into it but cannot promise anything.

bkueng:
Superseded by https://github.com/vladbat00/bevy_egui/pull/479

PR #474, opened 2026-04-15
Fix egui 0.34 compatibility

freesig:
Bumps to egui 0.34 and fixes accesskit plus a few minor version changes

PR #469, opened 2026-03-14
Bevy 0.19 + Egui 0.34

taboky-dev:
:warning: **This is still a WIP, I am still testing and fixing some stuff, I will be updating the broken and todo sections as I work on this.**

To use this branch, add this to your `Cargo.toml`: 

```toml
[dependencies]
bevy = "0.19.0-rc.2"
bevy_egui = { git = "https://github.com/taboky-dev/bevy_egui.git", rev = "ee645af" }
```

---

### Broken
- ~~The `accesskit` feature won't compile due to egui's `accesskit` dependency being outdated (https://github.com/emilk/egui/pull/7850)~~
- Some examples are not working as intended (see below)

### Todo
- [x] Add missing documentation
- [x] Fix ui render order
- [x] Update `wgpu-types` to 0.29 
- [x] Update `egui` to 0.34
- [x] `EguiAboveBevyUi` causes bevy ui to be un-pickable
- [x] Replace deprecated method `egui::Context::run` with `egui::Context::run_ui`
- [x] Fix broken examples
  - [x] `color_test` "Render to image (Mesh)" option is not rendering at all
  - [x] `paint_callback` panics
  - [x] `render_to_image_widget` does not compile
  - [x] `side_panel` prints warnings every frame

(Closes #473)

PR #477, opened 2026-04-27
Update egui to `0.34`

apekros:
Updates egui to `0.34`, only small changes.

MSRV of egui was updated to 1.92, some functions deprecated (changed to equivalents) and accesskit feature no longer needed as egui always depends on it

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "bevy_egui"
 version = "0.39.1"
-rust-version = "1.89.0"
+rust-version = "1.92.0"
 authors = ["vladbat00 <vladyslav.batyrenko@gmail.com>"]
 description = "A plugin for Egui integration into Bevy"
 license = "MIT"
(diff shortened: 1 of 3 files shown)


vladbat00:
Hi, thanks! Looks good to me, let's just avoid formatting changes in `Cargo.toml`.

Upd. and let's fix those deprecation warnings :)

apekros:
> Hi, thanks! Looks good to me, let's just avoid formatting changes in `Cargo.toml`.
> 
> Upd. and let's fix those deprecation warnings :)

Apologies about the format changes didn't notice, thanks for the reply, will revert those, and fix the rest of those warnings :)

taboky-dev:
There was a PR (#474) that made similar changes, the accessibility feature causes a compilation error due to mismatched  `accesskit` versions (no longer optional in `egui`).

In the PR #469 I mentioned the commit 0af8366 (which bumps `egui` to `0.34`) that also has similar changes but it is incomplete; it replaces only 1 of the 2 `run` methods to `run_ui` (same thing here).

That change (`run -> run_ui`) on its own introduced a bug in that PR; it makes Bevy UI un-interactable/pickable if `egui` is being rendered above it, but reverting `run_ui` back to `run` causes another bug regarding text selection in `egui` (non-stop selection/unable to deselect text).

Neither of them were fixed in #469 because I didn't have much time to look into it, I just reverted `run_ui` back to `run` temporarily and ignored the selection bug.

If the accessibility feature being broken isn't a blocking thing then test if Bevy UI is usable, this issue could very well be something that happens bevy `0.19` for some reason, or caused by one of my previous changes.

apekros:
> There was a PR (#474) that made similar changes, the accessibility feature causes a compilation error due to mismatched `accesskit` versions (no longer optional in `egui`).
> 
> In the PR #469 I mentioned the commit [0af8366](https://github.com/vladbat00/bevy_egui/commit/0af83660d2036a63ab0bd53d051aa6d5c3c8bcf6) (which bumps `egui` to `0.34`) that also has similar changes but it is incomplete; it replaces only 1 of the 2 `run` methods to `run_ui` (same thing here).
> 
> That change (`run -> run_ui`) on its own introduced a bug in that PR; it makes Bevy UI un-interactable/pickable if `egui` is being rendered above it, but reverting `run_ui` back to `run` causes another bug regarding text selection in `egui` (non-stop selection/unable to deselect text).
> 
> Neither of them were fixed in #469 because I didn't have much time to look into it, I just reverted `run_ui` back to `run` temporarily and ignored the selection bug.
> 
> If the accessibility feature being broken isn't a blocking thing then test if Bevy UI is usable, this issue could very well be something that happens bevy `0.19` for some reason, or caused by one of my previous changes.

The non-stop selection/unable to deselect text I believe is related to not running the plugins in egui after the run call

vladbat00:
Egui was updated to `0.34` as part of #469. But thank you for the PR anyway!

PR #479, opened 2026-05-19
clipboard: add smithay-clipboard for wayland support

bkueng:
As discussed in https://github.com/vladbat00/bevy_egui/pull/476.
This follows the implementation of winit.

Tested under Linux with and without 'wayland' feature.

The need to pass a window handle required a bit of a different initialization strategy.
And I'm not sure if there is a way or guarantee in bevy for the destruction order of resources: strictly speaking the Display needs to be destroyed after the smithay clipboard instance.


Friz64 (approved):
tested this, works fine. code looks good imo. thank you

vladbat00:
Looks great, ty for driving this!

PR #481, opened 2026-06-10
Bump bevy to `0.19.0-rc.3`

taboky-dev:


vladbat00:
If there are no changes required to make `bevy_egui` compile with `rc.3`, we can skip the bump, as cargo treats different rc versions as compatible (i.e. it won't pull both `rc.2` and `rc.3` because of `bevy_egui`).

taboky-dev:
Ah, you are right. I was having some mismatched version issue trying to update a game to bevy `rc.3` for some reason, but I see the issue came from somewhere else.

Oopsies, closing this.

PR #484, opened 2026-06-22
Schedule 3D egui pass after TAA

### LINKED ISSUE #483: bevy_egui 0.40: 3D egui pass can run before TAA
In a Bevy 0.19 app using `TemporalAntiAliasing`, I’m seeing egui contents get accumulated by TAA when rendered on a 3D camera. The visible result is old egui windows trailing/smearing across the frame.
Locally, this was fixed by scheduling the 3D egui pass after `Core3dSystems::EarlyPostProcess` instead of only after `Core3dSystems::MainPass`:

```rust
let egui_pass_3d = render::egui_pass
    .after(bevy_core_pipeline::Core3dSystems::EarlyPostProcess)
    .before(bevy_core_pipeline::upscaling::upscaling);
```

Not sure if `EarlyPostProcess` is the exact preferred ordering, but egui seems to need to run after TAA and before upscaling.

kurtkuehnert:
Fixes #483.

The 3D egui render pass was scheduled after `Core3dSystems::MainPass`, which means it could run before Bevy's early post-process systems such as TAA. When TAA is enabled, egui output could then be accumulated into the frame history, causing UI windows to smear or trail.

This schedules the 3D egui pass after `Core3dSystems::EarlyPostProcess` while still keeping it before upscaling, so egui renders after TAA and remains crisp.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1307,7 +1307,7 @@ impl Plugin for EguiPlugin {
                 .after(bevy_core_pipeline::Core2dSystems::MainPass)
                 .before(bevy_core_pipeline::upscaling::upscaling);
             let egui_pass_3d = render::egui_pass
-                .after(bevy_core_pipeline::Core3dSystems::MainPass)
+                .after(bevy_core_pipeline::Core3dSystems::EarlyPostProcess)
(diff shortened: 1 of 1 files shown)


vladbat00:
Great, thanks for the PR!

PR #486, opened 2026-06-24
Derive `Default` for `PrimaryEguiContext`

DavidCrossman:
The `PrimaryEguiContext` component currently cannot be used in BSN since it does not implement `FromTemplate`. Since `FromTemplate` is blanket implemented for all `Clone + Default` types, implementing `Default` for `PrimaryEguiContext` will enable its use in BSN.


vladbat00:
Thanks for the PR! Will merge and release a new patch version soon.

PR #491, opened 2026-07-09
fix: schedule egui 3d render pass after PostProcess to prevent tonemapping race

### LINKED ISSUE #490: Flickering with ToneMapping
I noticed on my Macbook, that I got flickering UI when I changed to a tonemapper like AgX.

pretty easy to reproduce locally on Metal with turning that on and off on the camera.

the fix might be a one-line change from `.after(EarlyPostProcess)` to `.after(PostProcess)` for egui_pass_3d?

g4borg:
Closes #490 

Tested also 2d with HDR+tonymcmapface, flickering did not affect 2d pipeline, left it .after(MainPass)


vladbat00:
Great, thank you for the fix!

PR #493, opened 2026-08-09
Update to egui 0.36

barsoosayque:
Pretty casual update to `egui = "0.36"`. One notable thing is that `DroppedFile` is now a trait, so *bevy_egui* now has its own implementation of that. Sadly, the new trait doesn't have any additional info.


vladbat00:
Looks great, thank you!

PR #496, opened 2026-08-11
Preserve cursor and mouse button event order

### LINKED ISSUE #495: PointerButton events get the end-of-frame cursor position instead of the position at the press
bevy_egui 0.41.1, bevy 0.19, macOS.

`write_pointer_button_messages_system` stamps every button event with `EguiContextPointerPosition`, which by that point holds the last cursor position of the frame rather than the position the pointer was at when the button actually changed state.

The system ordering makes it explicit. `write_window_pointer_moved_messages_system` is in `EguiInputSet::InitReading` and drains every `CursorMoved` for the frame, leaving `EguiContextPointerPosition` sitting at the last one. `write_pointer_button_messages_system` is in `EguiInputSet::FocusContext`, which is chained strictly after it, and then does:

```rust
egui_input_message_writer.write(EguiInputEvent {
    context,
    event: egui::Event::PointerButton {
        pos: context_pointer_position.position,
        button,
        pressed,
        modifiers,
    },
});
```

A mouse emits pointer events much faster than an app renders frames, so a single frame's batch is typically something like move, move, press, move, move. The press therefore reaches egui carrying whatever position the pointer had got to by the end of that batch. The error is one frame of pointer travel, so it scales with frame time and with how fast the pointer was moving when the button went down. At 60fps and a brisk drag that is easily tens of pixels, which is enough to put the press outside the rect of whatever you were aiming at.

There is a second effect from the same cause. Because the moved system runs first and drains everything, egui receives the frame as "every move, and then the button", so the press is out of sequence as well as out of position. That matters beyond the raw coordinate: `InputState` only updates `has_moved_too_much_for_a_click` on `PointerMoved` events that arrive after `press_origin` has been set, so ordering feeds into click versus drag disambiguation too.

The information is not lost. Bevy keeps the events interleaved in true arrival order in `Messages<WindowEvent>`, written as a single `write_batch` in `bevy_winit/src/state.rs`. Walking that one stream and emitting `PointerMoved` and `PointerButton` as they come through gives egui both the correct position and the correct order.

I have done this in my own app by turning both systems off through `EguiInputSystemSettings` and writing a replacement, which fixed presses landing measurably away from where I had clicked. Happy to open a PR if you would like it upstream. The part that needs some care is that the button path uses `read_with_non_window_hovered` while the moved path uses plain `read`, so merging the two has to preserve the non window context handling. I did not need any of that locally, so my version skips it.

kevthedawg:
Fixes #495

Reads cursor movement and mouse button messages from `WindowEvent` so Egui receives them in their original order. This makes pointer button messages use the pointer position at the time of the button input instead of the final position for the frame.

The existing global and per-context input settings and non-window context handling are preserved.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -19,7 +19,7 @@ use bevy_input::{
 };
 use bevy_log::{self as log};
 use bevy_time::{Real, Time};
-use bevy_window::{CursorMoved, FileDragAndDrop, Ime, Window};
+use bevy_window::{CursorMoved, FileDragAndDrop, Ime, Window, WindowEvent};
 use bevy_winit::WinitUserEvent;
 use egui::{Modifiers, TouchPhase, ViewportId};
 
@@ -84,7 +84,7 @@ pub struct HoveredNonWindowEguiContext(pub Entity);
 /// starts exposing triangle index or UV.
 ///
 /// Updating focused contexts happens during [`crate::EguiInputSet::FocusContext`],
-/// see [`write_pointer_button_messages_system`] and [`write_window_touch_messages_system`].
+/// see [`write_pointer_moved_and_button_messages_system`] and [`write_window_touch_messages_system`].
 #[derive(Resource, Clone)]
 pub struct FocusedNonWindowEguiContext(pub Entity);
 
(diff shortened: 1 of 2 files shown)


vladbat00:
Great stuff, thank you! I'll have another closer look a bit later. One thing I can suggest already is to straight remove the systems that are marked as deprecated. I don't think a lot of people actually use those manually, and I don't want to forget to remove them in the future versions.

